PR #2, opened 2020-11-18
[JIRA] TH2-1027 Show request error code instead list

VekesserVladimir:


src/stores/EventsStore.ts
```diff
@@ -99,6 +99,8 @@ export default class EventsStore {
 
 	@observable isExpandedMap: Map<string, boolean> = new Map();
 
+	@observable errorStatusCode: number | null = null;
```
rzvdv, line 102: Can we rename it to eventTreeStatusCode to be more specific

src/stores/EventsStore.ts
```diff
@@ -220,6 +223,7 @@ export default class EventsStore {
 				this.eventTree = sortEventsByTimestamp(rootEventIds);
 			});
 		} catch (error) {
+			this.errorStatusCode = error.status;
```
rzvdv, line 226: error might be an AbortError

src/stores/MessagesStore.ts
```diff
@@ -83,6 +83,8 @@ export default class MessagesStore {
 	@observable
 	public scrollTopMessageId: string | null = null;
 
+	@observable errorStatusCode: number | null = null;
```
rzvdv, line 86: rename to messagesListErrorStatusCode code

rzvdv (changes requested):

rzvdv (approved):

PR #6, opened 2020-11-18
TH2-991. Ctrl + F triggers search for all of the tabs

VitalyMashanov:


src/components/event/EventWindow.tsx
```diff
@@ -22,12 +22,34 @@ import '../../styles/events.scss';
...
+	setIsActive: () => void;
+}
+
+const EventWindow = (props: EventWindowProps) => {
 	const viewStore = useEventWindowViewStore();
+	const { isActive, setIsActive } = props;
+	const ref = React.useRef<HTMLDivElement | null>(null);
+
+	const documentClickHandler = (e: MouseEvent) => {
+		const tabs = ref.current && ref.current.closest('.tabs');
```
rzvdv, line 36: if you do something jquery style that means something went wrong. I think we should move this click listener to AppWindow actually. Because in this component we dont have access to tabs. We also can check if tab is already active then we don't need to set it active again.

src/stores/AppWindowStore.ts
```diff
@@ -33,6 +33,10 @@ export default class AppWindowStore {
 		return this.tabs.length === 0;
 	}
 
+	@computed get activeTab() {
+		return this.tabs[this.activeTabIndex];
```
rzvdv, line 37: Let's add activeTabs computed property inside WindowsStore.

src/components/AppWindow.tsx
```diff
@@ -35,13 +35,28 @@ import { useWindowsStore } from '../hooks/useWindowsStore';
...
+	setActiveEventsTab: (eventsTab: EventsTab | null) => void;
 }
 
 const TabsWithSideDropTargets = withSideDropTargets(Tabs);
 
 const AppWindow = (props: AppWindowProps) => {
-	const { windowStore, windowIndex } = props;
+	const { windowStore, windowIndex, activeEventsTab, setActiveEventsTab } = props;
 	const windowsStore = useWindowsStore();
+	const [isActive, setIsActive] = React.useState(false);
```
rzvdv, line 47: we don't need that

src/components/AppWindow.tsx
```diff
@@ -128,7 +143,7 @@ const AppWindow = (props: AppWindowProps) => {
 			tabPanels={windowStore.tabs.map(tab =>
 				isEventsTab(tab) ? (
 					<EventWindowProvider value={tab.store}>
-						<EventWindow />
+						<EventWindow isActive={isActive} setIsActive={setActiveWindow} />
```
rzvdv, line 146: `isActive={tab === activeEventsTab}`

src/components/AppWindow.tsx
```diff
@@ -35,13 +35,28 @@ import { useWindowsStore } from '../hooks/useWindowsStore';
...
-	const { windowStore, windowIndex } = props;
+	const { windowStore, windowIndex, activeEventsTab, setActiveEventsTab } = props;
 	const windowsStore = useWindowsStore();
+	const [isActive, setIsActive] = React.useState(false);
+
+	const setActiveWindow = () => {
+		setActiveEventsTab(isEventsTab(windowStore.activeTab) ? windowStore.activeTab : null);
+	};
+
+	React.useEffect(() => {
```
rzvdv, line 53: i thought we have a reaction here that would look something like
```
React.useEffect(() => {
 if(!activeEventsTab || !windowsStore.activeTabs.contains(activeEventsTab)) {
    activeTab = windowsStore.activeTabs.find(isEventsTab) || null
 }
}, [windowsStore.activeTabs])
```

rzvdv (approved):

PR #14, opened 2020-11-26
TH2-1066. Implement "go to timestamp" input field, that accepts message, event, or timestamp.

VitalyMashanov:


src/components/util/TimestampInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+			if (!message) {
+				return;
+			}
+			const timestamp = getTimestamp(message.timestamp);
+			setCurrentValue(formatTimestampValue(timestamp, DateTimeMask.DATE_TIME_MASK));
+		});
+	};
+
+	const getTimestampFromEventOrMessage = (eventIdOrMessageId: string) => {
+		eventHttpApi.getEvent(eventIdOrMessageId, undefined, { probe: true }).then(event => {
```
rzvdv, line 69: let use async await and we definitely need an AbortController for both requests

rzvdv (changes requested):

PR #17, opened 2020-11-27
TH2-1066. Implement "go to timestamp" input field, that accepts message, event, or timestamp.

VitalyMashanov:


src/components/util/TimestampInput.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+	useOutsideClickListener(wrapperRef, outsideClickListener);
+
+	React.useEffect(() => {
+		const date = currentValue.split(' ')[0];
+		const eventAbortController = new AbortController();
+		const messageAbortController = new AbortController();
+
+		if (!focus && currentValue !== '' && !dateFormatPattern.test(date)) {
+			const event = getEvent(currentValue, eventAbortController.signal);
```
rzvdv, line 72: can we move all this logic into a function ? and delete getEvent/getMessage they don't seem to be useful

rzvdv (changes requested):

src/components/util/TimestampInput.tsx
```diff
@@ -0,0 +1,154 @@
...
+		<div className={wrapperClassName} ref={wrapperRef}>
+			<input {...inputProps} />
+			{datepicker && <button className={datepickerButtonClassName} onClick={toggleDatepicker} />}
+		</div>
+	);
+};
+
+TimestampInput.displayName = 'TimestampInput';
+
+export default TimestampInput;
```
rzvdv, line 154: This input isnt used anywhere. Add it to graph component as it on sketch

src/components/util/TimestampInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+		placeholder = 'Go to ID or Timestamp',
+		style,
+		onSubmit,
+	} = props;
+
+	const [currentValue, setCurrentValue] = React.useState('');
+	const [currentTimestamp, setCurrentTimestamp] = React.useState(0);
+	const [showPicker, setShowPicker] = React.useState(false);
+
+	const ac = new AbortController();
```
rzvdv, line 64: This should be inside of useEffect. You can pass abort signal to setTimestampFromEventOrMessage as argument

If one requests succeeds i think it's not a big deal to leave it hanging. It's only important we cancel previous requests on value change.

src/styles/timestamp-input.scss
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+ @import './common/vars';
+
+.timestamp-input {
+	position: relative;
+  &__input {
+		width: 100%;
```
rzvdv, line 22: make inputs width as it on sketch

src/styles/timestamp-input.scss
```diff
@@ -0,0 +1,49 @@
...
+		border: 1px solid $timestampInputBorderColor;
+		border-radius: 4px;
+		font-size: 10px;
+		outline: none;
+		transition: box-shadow ease-in-out 320ms;
+		box-shadow: none;
+		&::placeholder {
+			font-style: italic;
+		}
+		&:focus {
```
rzvdv, line 34: i don't think we need that

rzvdv (changes requested):

src/styles/timestamp-input.scss
```diff
@@ -0,0 +1,49 @@
...
+ ******************************************************************************/
+
+ @import './common/vars';
+
+.timestamp-input {
+	position: relative;
+  &__input {
+		width: 100%;
+		box-sizing: border-box;
+		padding: 5px 9px;
```
rzvdv, line 24: padding right need to be more than 9px so text doesn't overlap button

rzvdv (approved):

PR #19, opened 2020-12-01
TH2-1108. Make pinned messages persistent by using local browser storage.

VitalyMashanov:
* raw solution


src/stores/SelectedStore.ts
```diff
@@ -74,10 +79,27 @@ export class SelectedStore {
 
 	@action
 	public toggleMessagePin = (message: EventMessage) => {
+		const stringPersistedPinnedMessages = localStorage.getItem('pinnedMessages');
 		if (this.pinnedMessages.findIndex(m => m.messageId === message.messageId) === -1) {
+			if (stringPersistedPinnedMessages) {
```
rzvdv, line 84: just save this.pinnedMessages to localStorage after if-else block

src/stores/WindowsStore.ts
```diff
@@ -128,6 +128,11 @@ export default class WindowsStore {
 
 	@action
 	private init(initialState: WindowsUrlState | null) {
+		const persistedPinnedMessages = localStorage.getItem('pinnedMessages');
```
rzvdv, line 131: That's not necessary and shouldn't be here

rzvdv (changes requested):

src/util/LocalStorageWorker.ts
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************** */
+
+import { EventMessage } from '../models/EventMessage';
+
+class LocalStorageWorker {
+	getPersistedPinnedMessages() {
+		const stringPersistedPinnedMessages = localStorage.getItem('pinnedMessages');
+		return stringPersistedPinnedMessages ? JSON.parse(stringPersistedPinnedMessages) : [];
```
rzvdv, line 22: json parsing should be handled inside of try-catch block. also it's return type is any now. we should validate if it's message

PR #20, opened 2020-12-01
[JIRA] TH2-1079 Expanding attached items

VekesserVladimir:


src/helpers/chunk.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import moment from 'moment';
+import { EventMessage } from '../models/EventMessage';
+import { Chunk } from '../models/graph';
+import { getTimestampAsNumber, toUTC } from './date';
+
+export function filterListByChunkRange(chunk: Chunk, list: EventMessage[]) {
+	return list.filter(item =>
+		toUTC(moment(getTimestampAsNumber(item.timestamp))).isBetween(
+			moment(chunk.from),
```
rzvdv, line 25: can we pass just a timestamp without conerting it to a moment value?

src/components/graph/Graph.tsx
```diff
@@ -70,6 +75,21 @@ function Graph() {
 				.valueOf(),
 		);
 
+		const attachedItems: AttachedItem[] = filterListByChunkRange(
```
rzvdv, line 78: we can make it shorter
```
const attachedItems: AttachedItem[] = filterListByChunkRange(
			chunk,
			selectedStore.attachedMessages.concat(selectedStore.pinnedMessages),
		).map(message => ({
			value: message,
			type: graphStore.attachedMessages.includes(messages) ? 'attached' : 'pinned'
		}));
```

src/components/graph/GraphChunk.tsx
```diff
@@ -88,16 +106,67 @@ const GraphChunk: React.ForwardRefRenderFunction<HTMLDivElement, Props> = (props
 		return ticksArr;
 	}, [chunk]);
 
+	const attachedItemTimeSize = React.useMemo(() => {
```
rzvdv, line 109: thats a simple calculation that doesn't need to be momoized

src/components/graph/GraphAttachedItemGroup.tsx
```diff
@@ -0,0 +1,192 @@
...
+				top: 0,
+			};
+		}
+
+		const { left, top } = itemRef.current.getBoundingClientRect();
+
+		return {
+			left,
+			top,
+		};
```
rzvdv, line 60: having itemRef.current?.getBoundingClientRect() in dependecies makes no sence, this wont trigger rerender and it always returns a new object that will never equal previous

rzvdv (changes requested):

rzvdv (approved):

PR #25, opened 2020-12-04
[JIRA] TH2-1132 Implement bookmarks list

VekesserVladimir:


src/components/BookmarksPanel.tsx
```diff
@@ -0,0 +1,116 @@
...
+interface SavedItem {
+	value: EventMessage | EventAction;
+	type: 'pinned-message' | 'event';
+}
+
+const BookmarksPanel = () => {
+	const selectedStore = useSelectedStore();
+	const savedItems: SavedItem[] = React.useMemo(() => {
+		return ([...selectedStore.attachedMessages, ...selectedStore.pinnedMessages] as (
+			| EventAction
```
rzvdv, line 36: why do we need this type casting?

src/components/BookmarksPanel.tsx
```diff
@@ -0,0 +1,116 @@
...
+			| EventMessage
+		)[])
+			.map(item => ({
+				value: item,
+				type: isEventMessage(item) ? ('pinned-message' as 'pinned-message') : ('event' as 'event'),
```
rzvdv, line 41: make a enum for selected items types

src/components/BookmarksPanel.tsx
```diff
@@ -0,0 +1,116 @@
...
+			}))
+			.sort((firstItem, secondItem) => {
+				const getItemTimestamp = (item: SavedItem) =>
+					getTimestampAsNumber(
+						item.type === 'event'
+							? (item.value as EventAction).startTimestamp
```
rzvdv, line 47: we have type guards for Events and Message, (isEventMessage, isEventAction) why don't we check value? either use it or create a new type guard based on item type

src/components/BookmarksPanel.tsx
```diff
@@ -0,0 +1,116 @@
...
+					item.type === 'event'
+						? (item.value as EventAction).successful
+							? 'passed'
+							: 'failed'
+						: null,
+				);
+				return (
+					<div
+						key={
+							item.type === 'event'
```
rzvdv, line 82: same

src/stores/SelectedStore.ts
```diff
@@ -77,6 +77,14 @@ export class SelectedStore {
 		localStorageWorker.setPersistedPinnedMessages(this.pinnedMessages);
 	};
 
+	@action
+	public removePinnedMessageFromLocalStorage = (removedMessage: EventMessage) => {
```
rzvdv, line 81: maybe we'll just have a reaction on pinnedMessagesChanges which updaes pinned messages to localStorage?

src/util/LocalStorageWorker.ts
```diff
@@ -16,18 +16,36 @@
 
 import { EventMessage } from '../models/EventMessage';
 
+const persistedPinnedMessageKeyPart = 'pinnedMessage';
 class LocalStorageWorker {
 	getPersistedPinnedMessages(): EventMessage[] {
 		try {
-			const stringPersistedPinnedMessages = localStorage.getItem('pinnedMessages');
-			return stringPersistedPinnedMessages ? JSON.parse(stringPersistedPinnedMessages) : [];
+			const persistedPinnedMessages: EventMessage[] = [];
```
rzvdv, line 23: i don't like idea of keeping every pinned message separately, why don't we just store them inside of collection?

rzvdv (changes requested):

src/components/BookmarksPanel.tsx
VekesserVladimir: We need this casting until we get pinned messages functionality. It shows that the array contains events and messages

rzvdv: i still don't get it. either way you can create observable pinnedEvents, initialize it with empty array and add functionality later

rzvdv (approved):

PR #27, opened 2020-12-08
TH2-1133. Implement SSE client

VitalyMashanov:


YourWeebyNeighbor:
I would implement the client without using the mobx store. There's going to be only one listener per request, so no need to make a dedicated worker for this.

PR #29, opened 2021-01-11
Th2 1133 Implement SSE client

VitalyMashanov:


src/components/search-panel/SearchPanel.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+		setCurrentlyLaunchedChannel(channel);
+	}, [formType, formState, eventFilter, messagesFilter]);
+
+	const stopChannel = useCallback(() => {
+		currentlyLaunchedChannel?.close();
+		setCurrentlyLaunchedChannel(null);
+	}, [currentlyLaunchedChannel]);
+
+	const currentResults: Result[] = React.useMemo(() => {
```
rzvdv, line 182: I dont think it very usefull to reiterate over values just to add a type and then inside of SearchPanelResults you checking type again. Just pass results as they are

src/components/search-panel/SearchPanelFilters.tsx
```diff
@@ -0,0 +1,165 @@
...
+	const [currentValues, setCurrentValues] = useSetState<Values>({
+		type: '',
+		body: '',
+		name: '',
+		attachedEventIds: '',
+	});
+
+	const setCurrentValue = (name: string) => (value: string) => {
+		setCurrentValues(currentState => ({
+			...currentState,
```
rzvdv, line 106: So what is the point of useSetState hook? you use it just like useState

src/components/search-panel/SearchPanelForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+	inputConfig: DateTimeInputType;
+};
+
+const SearchPanelForm = (props: SearchPanelFormProps) => {
+	const { formType, state, setState } = props;
+	const [currentStream, setCurrentStream] = useState('');
+
+	const toggleSearchDirection = () => {
+		setState(currentState => ({
+			...currentState,
```
rzvdv, line 46: same. i thought you used useSetState hook to be able to set state in patches but you stil spread over previous state as you would do with useState hook

src/components/search-panel/SearchPanelForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+		id: 'stream',
+		label: 'Stream',
+		values: state.stream,
+		setValues: getFormStateUpdater('stream'),
+		currentValue: currentStream,
+		setCurrentValue: setCurrentStream,
+		autocompleteList: null,
+	};
+
+	const config: FilterRowConfig[] = useMemo(() => {
```
rzvdv, line 102: every render commonFormConfig will be a new variable so useMemo is useless here

rzvdv (changes requested):

PR #32, opened 2021-01-26
TH2-1386. Implement search history

VitalyMashanov:


rzvdv:
Replace  `queryParams?: Record<string, string | number | boolean | null | string[]>` inside of ApiSchema to parameters that api accepts.

rzvdv:
fix return type of `getFilters: (filterType: 'events' | 'messages') => Promise<[string]>;` to `Promise<string[]>`

src/components/search-panel/SearchPanel.tsx
```diff
@@ -42,6 +44,62 @@ export type SearchPanelState = {
 	stream: string[];
 };
 
+export type Result = EventTreeNode | EventMessage;
+
+export type SearchHistory = {
```
rzvdv, line 49: Maybe we should store requests and results together ? e.g. `{ request: SearchRequest, result: SearchResult, id: .... }`

src/components/search-panel/SearchPanel.tsx
```diff
@@ -128,12 +251,6 @@ const SearchPanel = () => {
 
 		const currentFilters = formType === 'event' ? eventFilterInfo : messagesFilterInfo;
```
rzvdv, line 252: You have memoized value `filter`. so we can use filter.info

src/components/search-panel/SearchPanel.tsx
```diff
@@ -128,12 +251,6 @@ const SearchPanel = () => {
...
 
-		const filterParams = formType === 'event' ? eventFilter : messagesFilter;
-
-		function getFilter<T extends keyof FilterState>(name: T) {
-			return filterParams[name];
-		}
-
 		const filtersToAdd = currentFilters
 			.filter((info: SSEFilterInfo) => {
 				const values = getFilter(info.name).values;
```
rzvdv, line 256: can be shortened into 1 line

rzvdv:
I dont think sse api `getEventSource` should accept listeners or any callbacks, it should just return EventSource

src/components/search-panel/SearchPanel.tsx
```diff
@@ -191,7 +322,7 @@ const SearchPanel = () => {
 					config={{
```
rzvdv, line 322: There is a lot of state here, so everything is rerendering, we can start optimizing by memoizing toggler config

PR #45, opened 2021-02-10
TH2 1439. Add history to the "go to" component

VitalyMashanov:


rzvdv:
Let's rename TimestampInput and all related to this component to GraphSearch. And move it to graph folder. I doubt this component will ever be used outside of graph.

And about history. let's have a local state inside GraphSearch component that initializes with localStorage, and inside useEffect hook we can save history on change

rzvdv (changes requested):
Let's rename TimestampInput and all of it's relative components to GraphSearch.

Let's store searchHistory inside state, whice will be initialized with localStorage, and using useEffect we will keep localStorage sync

src/components/graph/GraphSearchHistory.tsx
```diff
@@ -0,0 +1,61 @@
...
+			{history.length !== 0 && (
+				<>
+					<p>history</p>
+					<hr />
+					{history.map(item => (
+						<BookmarkItem
+							key={computeKey(item)}
+							item={item}
+							onClick={activeWorkspace.onSavedItemSelect}
+							onRemove={() => {
```
VekesserVladimir, line 48: This function will be the reason of wasted renders. Place this function above with useCallback hook

src/components/graph/GraphSearchHistory.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const GraphSearchHistory = (props: Props) => {
+	const { history } = props;
+	const activeWorkspace = useActiveWorkspace();
+
+	const computeKey = (item: EventAction | EventMessage) => {
+		return isEventAction(item) ? item.eventId : item.messageId;
+	};
+
+	const renderHistory = React.useCallback(() => {
```
rzvdv, line 39: Let's not memoize rendering function, observer components are already memoized. Also be careful about dependencies when using useCallback, useMemo. right now `activeWorkspace.onSavedItemSelect` could change but it will not rerender because it's not in dependency list.
Instead we can memoize onRemove function, but that would be useless because currently BookmarkItem isn't memoized. So lets use React.memo in BookmarkItem component

also we can simplify a bit rendering

if (!length) return something

return history items

rzvdv (approved):

PR #49, opened 2021-02-11
Highlight for messages and events.

artem-davidovich:


src/components/message/message-card/MessageCard.tsx
```diff
@@ -70,6 +71,7 @@ function MessageCardBase({ message, showRaw, showRawHandler }: Props) {
 		isSelected ? 'selected' : null,
 		isAttached ? 'attached' : null,
 		isPinned ? 'pinned' : null,
+		isHighlighted ? 'highlighted' : null,
```
rzvdv, line 74: I thought we would have some effect that will reset hightlightedMessageId after applying highlight styles to it, because if we will not reset it might highlight again

src/stores/messages/MessagesStore.ts
```diff
@@ -440,6 +445,7 @@ export default class MessagesStore {
 		const messageIndex = this.messagesIds.indexOf(messageId);
 		if (messageIndex !== -1) {
 			this.scrolledIndex = new Number(messageIndex);
+			this.highlightedMessageId = messageId;
```
rzvdv, line 448: that's not a place to set highlighted message, instead we should do it in `onSavedItemSelect`

src/components/message/message-card/MessageCard.tsx
artem-davidovich: Fixed

src/stores/messages/MessagesStore.ts
artem-davidovich: Fixed

rzvdv (approved):

PR #51, opened 2021-02-15
search-panel-improvements.

VitalyMashanov:
* autocomplete list for sessions
 * toast messages for sse errors
 * text-overflow mixin

### DIFF
--- a/src/components/search-panel/SearchPanel.tsx
+++ b/src/components/search-panel/SearchPanel.tsx
@@ -33,11 +33,6 @@ const SearchPanel = () => {
 	const { ref: searchPanelRef } = useActivePanel(null);
 	const searchStore = useSearchStore();
 	const messagesStore = useMessagesWorkspaceStore();
-	const {
-		filterStore: {
-			messagesFilter: { streams },
-		},
-	} = messagesStore;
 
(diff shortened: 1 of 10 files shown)


src/components/search-panel/SearchPanel.tsx
```diff
@@ -32,6 +32,12 @@ const SearchPanel = () => {
 	const activeWorkspace = useActiveWorkspace();
 	const { ref: searchPanelRef } = useActivePanel(null);
 	const searchStore = useSearchStore();
+	const messagesStore = useMessagesWorkspaceStore();
+	const {
+		filterStore: {
+			messagesFilter: { streams },
```
rzvdv, line 38: i think you should have taken messageSessions from messages store for autocomplete list

src/stores/SearchStore.ts
```diff
@@ -319,6 +320,20 @@ export class SearchStore {
 		localStorageWorker.saveSearchHistory(this.searchHistory);
 	};
 
+	@action
```
VekesserVladimir, line 323: We doesn't need action decorator here

rzvdv (approved):

PR #55, opened 2021-02-16
Show hovered events and messages on graph

VitalyMashanov:
* messages\events are shown in graph on message\event hover


src/components/event/EventCardHeader.tsx
```diff
@@ -83,8 +84,18 @@ function EventCardHeader({
...
 
 	return (
-		<div className={rootClassName} onClick={onSelect} style={rootStyle}>
-			<div className={iconClassName}></div>
+		<div
+			className={rootClassName}
+			onClick={onSelect}
+			style={rootStyle}
+			onMouseEnter={() => {
+				eventStore.setHoveredEvent(event);
```
rzvdv, line 92: we definitely need some debouncing for this, otherwise it will be called to ofter and often unnecessary

src/stores/events/EventsStore.ts
```diff
@@ -80,6 +80,9 @@ export default class EventsStore {
 
 	@observable.ref selectedNode: EventTreeNode | null = null;
 
+	@observable
```
rzvdv, line 83: let make it observable.ref, event properties won't change and we dont need to observe them

src/components/graph/GraphItemsGroup.tsx
```diff
@@ -20,21 +20,24 @@ import { createBemElement, createStyleSelector } from '../../helpers/styleCreato
...
 	| EventStatus.PASSED;
 
 const listIconsPriority: { [key in GroupItemType]: number } = {
-	[GraphItemType.ATTACHED_MESSAGE]: 3,
-	[GraphItemType.PINNED_MESSAGE]: 2,
-	[EventStatus.FAILED]: 1,
-	[EventStatus.PASSED]: 1,
+	[GraphItemType.ATTACHED_MESSAGE]: 4,
+	[GraphItemType.PINNED_MESSAGE]: 3,
+	[GraphItemType.HOVERED_EVENT]: 1,
```
rzvdv, line 37: I think hovered events and messages should have maximum priority

src/components/event/EventCardHeader.tsx
```diff
@@ -82,9 +83,23 @@ function EventCardHeader({
 		e.stopPropagation();
 	};
 
+	const hoverEvent = useDebouncedCallback(() => {
```
rzvdv, line 86: This way hoverEvent could be triggered even when card isn't hovered.

create ref with timeout. onMouseOver assign ref a timeout with delay. onMouseLeave clear timeout.
I did something similar in TimeSelector component inside GraphChunksVirtualizer file

rzvdv (changes requested):
GraphItemsMenu event icons are still broken due to `getGraphItemType` function change

rzvdv (approved):

PR #57, opened 2021-02-18
TH2-1517 Use sse search api to render events

VekesserVladimir:


rzvdv:
Switching to flat view breaks app
Initial loading with event path doesn't expand events

src/stores/events/EventsStore.ts
```diff
@@ -181,41 +185,115 @@ export default class EventsStore {
...
+		return new Promise<void>((res, rej) => {
+			this.eventTreeStatusCode = null;
+			this.selectedNode = null;
+			this.isLoadingRootEvents = true;
+			this.eventTree = [];
 
-		this.selectedNode = null;
-		this.isLoadingRootEvents = true;
+			try {
+				const eventSource = this.api.sse.getEventSource({
```
rzvdv, line 197: sse channel should be member of eventstore and it should be close upon applying filter etc

src/stores/events/EventsStore.ts
```diff
@@ -181,41 +185,115 @@ export default class EventsStore {
...
-	private eventTreeAC: AbortController | null = null;
-
 	@action
-	private fetchEventTree = async () => {
-		if (this.eventTreeAC) {
-			this.eventTreeAC.abort();
-		}
-		this.eventTreeAC = new AbortController();
-		this.eventTreeStatusCode = null;
+	private fetchEventTree = () => {
```
rzvdv, line 189: store is getting quite large, maybe we can create some EventsDataStore to move everything related to fetching there?

src/stores/events/EventsStore.ts
```diff
@@ -181,41 +185,115 @@ export default class EventsStore {
...
+
+	@action
+	private async loadRootEvents() {
+		const loadEventById = async (eventId: string, childList: EventTreeNode[]) => {
+			try {
+				const event = await this.api.events.getEvent(eventId);
+				if (event.parentEventId === null) {
+					if (this.eventTree.find(node => node.eventId === event.eventId)) return;
+
+					const rootNode = convertEventActionToEventTreeNode(event);
```
rzvdv, line 276: Maybe converting is unnecessary ? if we receive EventAction let's make it a tree of EventActions?

YourWeebyNeighbor:
Let's not merge it just yet. We need to test how this change works on large amount of events

YourWeebyNeighbor (approved):

PR #61, opened 2021-02-24
Sse client refactoring  and search workspace extracting

VekesserVladimir:


src/components/graph/Graph.tsx
```diff
@@ -110,23 +111,41 @@ function Graph() {
...
-				setRange: activeWorkspace.eventsStore.onRangeChange,
-			},
-			{
-				type: 'messages-panel',
-				range: activeWorkspace.messagesStore.panelRange,
-				setRange: activeWorkspace.messagesStore.onRangeChange,
-			},
-		];
-	}, [activeWorkspace.eventsStore.panelRange, activeWorkspace.messagesStore.panelRange]);
+		return isWorkspaceStore(activeWorkspace)
```
rzvdv, line 114: I think a better way to solve this would be to have an intermediate component that returns just graph background if active workspace is SearchWorkspace and return Graph component otherwise

src/components/graph/GraphOverlay.tsx
```diff
@@ -28,6 +28,7 @@ import { getTimestampAsNumber } from '../../helpers/date';
 import { GraphStore } from '../../stores/GraphStore';
 import { OutsideItems, OutsideItemsMenu, OutsideItemsList } from './OutsideItems';
 import '../../styles/graph.scss';
+import { isWorkspaceStore } from '../../helpers/workspace';
```
rzvdv, line 31: unused

src/api/event.ts
```diff
@@ -25,7 +25,7 @@ const eventHttpApi: EventApiSchema = {
 			name: filter.names,
 			type: filter.eventTypes,
 		});
-		const res = await fetch(`backend/search/events?${params}`, { signal });
+		const res = await fetch(`search/events?${params}`, { signal });
```
rzvdv, line 28: this shouldn't be in master

src/components/message/MessagesWindowHeader.tsx
```diff
@@ -16,11 +16,12 @@
 
 import React from 'react';
 import { observer } from 'mobx-react-lite';
-import { useMessagesWorkspaceStore } from '../../hooks';
+import { useMessagesWorkspaceStore, useWorkspaceStore } from '../../hooks';
```
rzvdv, line 19: unused import

src/components/search-panel/SearchPanel.tsx
```diff
@@ -59,7 +58,7 @@ const SearchPanel = () => {
 						form={searchStore.searchForm}
 						formType={searchStore.formType}
 						updateForm={searchStore.updateForm}
-						streams={searchStore.formType === 'message' ? messagesStore.messageSessions : null}
+						streams={null}
```
rzvdv, line 61: Streams are required. They are being fetched in SearchStore now

src/helpers/url.ts
```diff
@@ -44,65 +45,73 @@ export function createURLSearchParams(
...
+				workspacesInitialStates: [
+					toJS({
+						events: eventStoreState,
+						messages: messagesStoreState,
+						timeRange: activeWorkspace.graphStore.range,
+						interval: activeWorkspace.graphStore.interval,
+						layout: activeWorkspace.viewStore.panelsLayout,
+					}),
+				],
+				searchWorkspaceInitialState: {
```
rzvdv, line 110: i dont think we need to store search url state. url is big enought already

rzvdv (changes requested):

src/components/graph/GraphBackground.tsx
```diff
@@ -0,0 +1,77 @@
...
+				type: 'messages-panel',
+				range: searchWorkspaceStore.graphStore.range,
+				setRange: searchWorkspaceStore.onRangeChange,
+			},
+		];
+	}, [searchWorkspaceStore.graphStore.range]);
+
+	return (
+		<div className='graph' ref={rootRef}>
+			<GraphOverlay
```
rzvdv, line 66: `<GraphSearch />` should be also enabled when SearchWorkspace is active

src/components/graph/Graph.tsx
```diff
@@ -127,26 +131,24 @@ function Graph() {
...
+				renderChunk={renderChunk}
+				setRange={activeWorkspace.graphStore.setRange}
+				getChunk={getChunk}
+				interval={activeWorkspace.graphStore.interval}
+				timestamp={activeWorkspace.graphStore.timestamp}
+				key={activeWorkspace.graphStore.timestamp}
+				panelsRange={panelsRange}
+				range={activeWorkspace.graphStore.range}
+			/>
 			<GraphOverlay
```
rzvdv, line 146: I dont know if GraphOverlay component make sence now since all previous markup and components are gone. Maybe we should move outside utems components to Graph

src/stores/workspace/SearchWorkspaceStore.ts
```diff
@@ -0,0 +1,103 @@
...
+	constructor(
+		private workspacesStore: WorkspacesStore,
+		private selectedStore: SelectedStore,
+		api: ApiSchema,
+	) {
+		this.searchStore = new SearchStore(api);
+
+		this.viewStore = new WorkspaceViewStore(undefined);
+
+		this.graphStore = new GraphStore(selectedStore, null);
```
rzvdv, line 48: It seems like SearchWorkspace doesnt need graphStore since it always disabled

rzvdv (changes requested):

src/components/search-panel/SearchPanel.tsx
```diff
@@ -25,11 +25,12 @@ import SearchPanelResults from './SearchPanelResults';
 import SearchPanelProgressBar from './SearchPanelProgressBar';
 import { FilterRowTogglerConfig } from '../../models/filter/FilterInputs';
 import '../../styles/search-panel.scss';
+import SearchWorkspaceStore from '../../stores/workspace/SearchWorkspaceStore';
 
 export type SearchPanelType = 'event' | 'message';
 
 const SearchPanel = () => {
-	const activeWorkspace = useActiveWorkspace();
+	const activeWorkspace = useActiveWorkspace() as SearchWorkspaceStore;
```
rzvdv, line 33: we'll useSearchWorkspace here

src/components/graph/GraphBackground.tsx
```diff
@@ -0,0 +1,41 @@
...
+	searchWorkspaceStore: SearchWorkspaceStore;
+}
+
+function GraphBackground({ searchWorkspaceStore }: GraphBackgroundProps) {
+	const rootRef = React.useRef<HTMLDivElement>(null);
+
+	return (
+		<div className='graph' ref={rootRef}>
+			<i className='th2-logo' />
+			<GraphSearch
```
rzvdv, line 33: `GraphSearch` value resets after switching workspaces. i think it should be inside `GraphWrapper`

src/components/graph/GraphBackground.tsx
```diff
@@ -0,0 +1,41 @@
...
+import '../../styles/graph.scss';
+import SearchWorkspaceStore from '../../stores/workspace/SearchWorkspaceStore';
+import GraphSearch from './search/GraphSearch';
+
+interface GraphBackgroundProps {
+	searchWorkspaceStore: SearchWorkspaceStore;
+}
+
+function GraphBackground({ searchWorkspaceStore }: GraphBackgroundProps) {
+	const rootRef = React.useRef<HTMLDivElement>(null);
```
rzvdv, line 28: unused ref

src/components/graph/GraphBackground.tsx
rzvdv: Lets move all this markup to `GraphWrapper` we dont need this component.
Both SearchWorkspaceStore and WorkspaceStore have onTimestampSelect & onSavedItemSelect methods

src/components/graph/GraphWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+ ***************************************************************************** */
+
+import { observer } from 'mobx-react-lite';
+import * as React from 'react';
+import { isWorkspaceStore } from '../../helpers/workspace';
+import { useActiveWorkspace } from '../../hooks';
+import Graph from './Graph';
+import GraphBackground from './GraphBackground';
+
+const GraphWrapper = () => {
```
rzvdv, line 24: let's move this component to Graph and this component will be exporting by default. And then we import it as `Graph` in App.tsx. so there will be no confusion about what is wrapper doing

src/stores/SelectedStore.ts
```diff
@@ -42,17 +43,28 @@ export class SelectedStore {
...
 	@computed get hoveredMessage(): EventMessage | null {
-		return this.workspacesStore.activeWorkspace.messagesStore.hoveredMessage;
+		return isWorkspaceStore(this.workspacesStore.activeWorkspace)
+			? this.workspacesStore.activeWorkspace.messagesStore.hoveredMessage
+			: null;
 	}
 
 	@computed get graphItems(): Array<GraphItem> {
-		const items = [...this.savedItems, ...this.workspacesStore.activeWorkspace.attachedMessages];
+		const items = [
```
rzvdv, line 58: i suppose we can return empty array if active workspace is SearchWorkspace

rzvdv (changes requested):

rzvdv (approved):

PR #66, opened 2021-02-26
Indicated panel

artem-davidovich:


rzvdv:
This implementation makes possible situation when both panels are active which is not what we want i think.

I don't understand why do we need these timeout which resets active panel.

If we want to change active panel on range change we can do that inside `EventStore.onRangeChange` or inside reaction to `MessagesStore.panelRange`

artem-davidovich:
@rzvdv 

> This implementation makes possible situation when both panels are active which is not what we want i think.
> 
> I don't understand why do we need these timeout which resets active panel.
> 
> If we want to change active panel on range change we can do that inside `EventStore.onRangeChange` or inside reaction to `MessagesStore.panelRange`

The task was highlight panel on update. I decided not to add new colors, but to combine the task with the existing functionality. That is why situation when both panels are active is okay.

Or should I do it differently?

PR #68, opened 2021-03-01
workspace-link-getter

VitalyMashanov:


src/components/WorkspaceLinkGetter.tsx
```diff
@@ -0,0 +1,51 @@
...
+		}
+		return () => {
+			clearTimeout(timeout);
+		};
+	}, [disabled]);
+
+	return (
+		<button
+			className='workspace-link-getter'
+			disabled={disabled}
```
rzvdv, line 39: I dont understand why this should be disabled for 3 seconds, but ok

src/stores/RootStore.ts
```diff
@@ -37,7 +38,8 @@ export default class RootStore {
```
rzvdv, line 38: After successful navigation via url, url stays the same and never updates. This might be confusing. We should delete it as soon as we parsed it.

rzvdv (changes requested):

rzvdv (approved):

PR #69, opened 2021-03-02
Workspace link getter

VitalyMashanov:


YourWeebyNeighbor (dismissed):

src/stores/RootStore.ts
```diff
@@ -38,8 +37,7 @@ export default class RootStore {
 
 	parseUrlState = (): WorkspacesUrlState | null => {
 		try {
-			const lastSearchQuery = window.location.search || localStorageWorker.getLastSearchQuery();
-			const searchParams = new URLSearchParams(lastSearchQuery);
+			const searchParams = new URLSearchParams(window.location.search);
```
rzvdv, line 40: I thought that was intentional to save last active state. But since it wasn't. Url middleware is always updating needlessly. I think we should refactor it to some method which returns url state or computed property. We can call it something like `appState` or `getAppState` and inside `WorkspaceLinkGetter` we can transform state object into url params

rzvdv (approved):

PR #73, opened 2021-03-02
Filter filling by search form

VekesserVladimir:


src/stores/workspace/WorkspaceStore.ts
```diff
@@ -157,23 +157,63 @@ export default class WorkspaceStore {
...
 			this.workspacesStore.addWorkspace(newWorkspace);
-			return;
-		}
+			res();
+		}).then(() => {
+			if (isEventMessage(savedItem)) {
+				const filterStore = newWorkspace.messagesStore.filterStore;
+				filterStore.setMessagesFilter(
+					filterStore.messagesFilter,
+					this.searchStore.savedSseMessagesFilters || {
```
rzvdv, line 184: So you are saving last applied filters in search panel and initializing Messages or Events panel with them. Hovewer `onSavedItemSelect` is also invoked on BookmarkItem click or on Item that was found in GraphSearch. And there is no guarantee that those filters and item have connections. We might want to have a separate handler for SearchPanel items. And those should be initialized with currentSearch.request.filters since every request has its own saved filter

src/stores/workspace/WorkspaceStore.ts
```diff
@@ -157,23 +157,63 @@ export default class WorkspaceStore {
...
+				this.messagesStore.onMessageSelect(savedItem);
+			} else {
+				this.viewStore.activePanel = this.eventsStore;
+				this.eventsStore.onEventSelect(savedItem);
+			}
+			return;
+		}
+
+		let newWorkspace: WorkspaceStore;
+		new Promise<void>(res => {
```
rzvdv, line 172: This workaround just as by as my timeout in MessagesStore. Eventually workspace should accept sse filters and properly initialize itself, im working on it now so if this workaround works, its ok for now

src/stores/SearchStore.ts
```diff
@@ -261,6 +265,12 @@ export class SearchStore {
 		const filterParams = this.formType === 'event' ? this.eventsFilter : this.messagesFilter;
 		if (this.searchChannel || !filterParams) return;
 
+		if (this.formType === 'event') {
```
rzvdv, line 268: filters for currentSearch are saved in currentSearch.request.filters

rzvdv:
This pr will be merged as part of #74

PR #79, opened 2021-03-05
search-progress-sse-heartbeat-visualization.

VitalyMashanov:
* search progress bar displays "keep_alive" event's timestamps

### DIFF
--- a/src/stores/SearchStore.ts
+++ b/src/stores/SearchStore.ts
@@ -17,7 +17,7 @@
 import { action, autorun, computed, observable, reaction, runInAction } from 'mobx';
 import moment from 'moment';
 import ApiSchema from '../api/ApiSchema';
-import { SSEFilterInfo, SSEParams } from '../api/sse';
+import { SSEFilterInfo, SSEHeartbeat, SSEParams } from '../api/sse';
 import { SearchPanelType } from '../components/search-panel/SearchPanel';
 import {
 	EventFilterState,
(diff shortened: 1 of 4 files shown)


src/stores/SearchStore.ts
```diff
@@ -120,20 +120,18 @@ export class SearchStore {
...
-		const lastItem = this.currentSearch?.results[this.currentSearch.results.length - 1];
-		const lastItemTimestamp = lastItem
-			? getTimestampAsNumber(isEventNode(lastItem) ? lastItem.startTimestamp : lastItem.timestamp)
-			: null;
 		return {
 			startTimestamp: this.searchForm.startTimestamp,
 			endTimestamp: this.searchForm.endTimestamp,
-			currentPoint: lastItemTimestamp
-				? lastItemTimestamp - Number(this.searchForm.startTimestamp)
+			currentPoint: this.currentSearchTimestamp
```
rzvdv, line 126: we need to store this data in currentSearch, otherwise loading state stays the same event if you are switching history

src/stores/SearchStore.ts
```diff
@@ -375,12 +374,27 @@ export class SearchStore {
 	};
 
 	private onChannelResponse = (ev: Event) => {
+		this.setCurrentSearchTimestamp(ev);
```
rzvdv, line 377: currently we are parsing event twice, which is unnecessary. lets handle 'keep_alive' event here also
parse event, add type guards and act accordingly

rzvdv (changes requested):
You also moved SearchPanelProgressBar outside of div.search-panel. So class 'spinner' isn't applying to it

rzvdv (approved):

PR #83, opened 2021-03-05
Jumping over graph items fix

VekesserVladimir:
Fixed bug on jumping over graph items by the arrow button clicking

### DIFF
--- a/src/stores/messages/MessagesStore.ts
+++ b/src/stores/messages/MessagesStore.ts
@@ -197,33 +197,16 @@ export default class MessagesStore {
 
 	@action
 	public onMessageSelect = async (message: EventMessage) => {
-		const filter = this.filterStore.filter;
-		const sseFilter = this.filterStore.sseMessagesFilter;
-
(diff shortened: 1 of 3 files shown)


src/stores/messages/MessagesStore.ts
```diff
@@ -197,33 +197,16 @@ export default class MessagesStore {
...
-			this.selectedMessageId = new String(message.messageId);
-			this.highlightedMessageId = message.messageId;
-			this.graphStore.setTimestamp(getTimestampAsNumber(message.timestamp));
-			this.selectedMessage = null;
-			this.workspaceStore.viewStore.activePanel = this;
-		} else {
-			this.selectedMessage = message;
-			this.handleFilterHint(message);
-		}
+		this.filterStore.resetMessagesFilter({
```
rzvdv, line 200: It's not a bug.
If we select message with session which isn't in filter, or if other filters applied we show user a hint that `filter need to be updated` and after used approves only then we navigate to message

PR #84, opened 2021-03-05
Move search input to search panel like filter, add hover effect for window control buttons

icefleen:


rzvdv (changes requested):
Let's make `SearchInput` behave just like `MultipleStringFilterRow` so that values are placed on multiple rows.

icefleen:
> Let's make `SearchInput` behave just like `MultipleStringFilterRow` so that values are placed on multiple rows.

Done: https://github.com/th2-net/th2-rpt-viewer/pull/84/commits/93b75d54a29dea25ab15a5046e54c2d2f12b6f3e

rzvdv (approved):

PR #95, opened 2021-03-12
Hint on hover panel time picker

VekesserVladimir:


src/components/graph/search/GraphSearchInput.tsx
```diff
@@ -91,6 +95,24 @@ function GraphSearchInput(props: Props) {
...
 
+	React.useEffect(() => {
+		const mask = inputConfig.mask || DATE_TIME_MASK;
+		const placeholder = inputConfig.placeholder || DATE_TIME_PLACEHOLDER;
+
+		setInputConfig({
+			...inputConfig,
+			value: pointerTimestamp
+				? moment.utc(pointerTimestamp).format(mask)
+				: timestamp
```
rzvdv, line 106: i set timestamp whenever user starts to input a valid date, so if user types '12:' timestamp is set.
So after changing input value from its previous value to value from `pointerTimestamp` and back
value sets to 12:00:00.000, even though user hasnt actually typed it.
I suggest when `pointerTimestamp` changes from null to number we save previous state, and after in chages back to null
we restore it

rzvdv: I also noticed an error i made. In `handleInputValueChange` in last else block where user is typing id i don't reset timestamp
Please add `setTimestamp(null)` there

rzvdv (dismissed):

rzvdv (approved):

PR #100, opened 2021-03-15
[WIP] Ability to submit value on focus and click away datepicker in graph

VekesserVladimir:


src/components/filter/date-time-inputs/FilterDatetimePicker.tsx
```diff
@@ -66,6 +73,7 @@ const FilterDatetimePicker = ({
 				appliedDate = dateValue.utc().startOf('day');
 			}
```
rzvdv, line 74: it saves value only if date was changed, but ignores time change.

PR #106, opened 2021-03-16
TH2-1628. Render messages in ascii and json mode as a single line

VitalyMashanov:


VekesserVladimir:
Снимок
There is some problem with aligning

src/components/message/message-card/MessageCard.tsx
```diff
@@ -193,7 +194,64 @@ function MessageCardBase({ message, viewType, setViewType }: Props) {
 		messagesStore.setHoveredMessage(null);
 	};
 
+	const renderMessageInfo = () => {
```
VekesserVladimir, line 197: I think we can combine renderMessageInfo and renderInlineMessageInfo into one method that takes MessageViewType

VekesserVladimir (approved):

PR #107, opened 2021-03-16
TH2-1630 Stop loading messages in 15 minutes interval

VekesserVladimir:


src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -387,4 +422,30 @@ export default class MessagesDataProviderStore {
 
 		return message;
 	};
+
```
rzvdv, line 425: @action

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -280,6 +311,8 @@ export default class MessagesDataProviderStore {
 			query,
 			this.onNextSoftFilterChannelResponse,
 			this.onLoadingError,
+			// eslint-disable-next-line @typescript-eslint/no-empty-function
+			() => {},
```
rzvdv, line 315: lets make it optional otherwise this looks bad. i'd rather explicitly pass undefined or null if i dont need that function

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -38,6 +41,12 @@ export default class MessagesDataProviderStore {
 	@observable
 	public isBeginReached = false;
 
+	@observable
+	public noMatchingMessagesPrev = false;
```
rzvdv, line 45: We need to reset these values at some point

src/stores/messages/SSEChannel.ts
```diff
@@ -65,6 +70,7 @@ export class SSEChannel {
 		private queryParams: MessagesSSEParams,
 		private onResponse: (messages: EventMessage[]) => void,
 		private onError: (event: Event) => void,
+		private onKeepAliveResponce: (event: KeepAliveResponce) => void,
```
rzvdv, line 73: typo ResponSe

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -387,4 +422,30 @@ export default class MessagesDataProviderStore {
...
+			...this.messagesStore.filterStore.messsagesSSEConfig.queryParams,
+			startTimestamp:
+				direction === 'previous' ? this.prevLoadEndTimestamp : this.nextLoadEndTimestamp,
+			searchDirection: direction,
+		} as MessagesSSEParams;
+
+		if (direction === 'previous') {
+			this.noMatchingMessagesPrev = false;
+			this.startPreviousMessagesChannel(query);
+			await this.searchChannelPrev.loadAndSubscribe();
```
rzvdv, line 444: this will return messages into nothing. you should use subscribe method

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -227,12 +249,21 @@ export default class MessagesDataProviderStore {
...
 	@action
-	public startNextMessagesChannel = (query: MessagesSSEParams) => {
+	public startNextMessagesChannel = (query: MessagesSSEParams, interval?: number) => {
 		this.searchChannelNext = new SSEChannel(
 			this.messagesStore.filterStore.messsagesSSEConfig.type,
 			query,
 			this.onNextChannelResponse,
 			this.onLoadingError,
+			interval
+				? e => {
```
rzvdv, line 259: There could be running 2 parallel channels if softFilterIs applied. I think we can add keep_alive handlers there too.
For example if you stop prevChannel, prevChannel with filters could be running.
I thought maybe we can add some checks in keep_alive where we can check if channel without filters is running.
if it isn't we can check keep_alive timestamp with last fetched message. if its later or sooner depending on search direction we might want to stop it

rzvdv (changes requested):

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -38,6 +41,12 @@ export default class MessagesDataProviderStore {
 	@observable
 	public isBeginReached = false;
 
+	@observable
+	public noMatchingMessagesPrev = false;
+
+	@observable
+	public noMatchingMessagesNext = false;
```
rzvdv, line 48: all the values should be reset inside resetMessagesDataState

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -181,12 +198,24 @@ export default class MessagesDataProviderStore {
...
+		this.prevLoadEndTimestamp = null;
+
 		this.searchChannelPrev = new SSEChannel(
 			this.messagesStore.filterStore.messsagesSSEConfig.type,
 			query,
 			this.onPrevChannelResponse,
 			this.onLoadingError,
+			interval
+				? e => {
+						if (query.startTimestamp - e.timestamp > interval * 1000 * 60) {
```
rzvdv, line 212: actually we cant compare `query.startTimestamp` with keep_alive timestamp.
Because startTimestamp is initialized once when we start loading, then when we scroll and load more we change only `resumeFromId`.

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -274,22 +315,40 @@ export default class MessagesDataProviderStore {
...
 		this.softFilterChannelNext = new SSEChannel(
 			this.messagesStore.filterStore.messsagesSSEConfig.type,
 			query,
 			this.onNextSoftFilterChannelResponse,
 			this.onLoadingError,
+			interval
+				? e => {
+						if (e.timestamp - query.startTimestamp > interval * 1000 * 60) {
+							this.softFilterChannelNext?.stop();
+							this.noMatchingMessagesNext = true;
```
rzvdv, line 328: SoftFilteredChannels shouldn't trigger `noMatchingMessagesNext` change

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -387,4 +452,31 @@ export default class MessagesDataProviderStore {
...
+	@action
+	keepLoading = async (direction: 'next' | 'previous') => {
+		if (
+			this.messagesStore.filterStore.filter.streams.length === 0 ||
+			!this.searchChannelNext ||
+			!this.searchChannelPrev
+		)
+			return;
+
+		const query = {
```
rzvdv, line 465: this query includes filters right now whether ifSoftFilter applied or not, and here we might want to resume soft channels also if they are stopped

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -387,4 +452,31 @@ export default class MessagesDataProviderStore {
...
+			...this.messagesStore.filterStore.messsagesSSEConfig.queryParams,
+			startTimestamp:
+				direction === 'previous' ? this.prevLoadEndTimestamp : this.nextLoadEndTimestamp,
+			searchDirection: direction,
+		} as MessagesSSEParams;
+
+		if (direction === 'previous') {
+			this.noMatchingMessagesPrev = false;
+			this.startPreviousMessagesChannel(query);
```
rzvdv, line 474: You have optional argument interval, so if it wasnt passed keep_alive callback isnt passed to SSEChannel, is that intentional? we will never restrict loading range after user clicks resume?

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -274,22 +315,40 @@ export default class MessagesDataProviderStore {
...
-	private startPrevSoftFilterChannel = (query: MessagesSSEParams) => {
+	private startPrevSoftFilterChannel = (query: MessagesSSEParams, interval?: number) => {
 		this.softFilterChannelPrev = new SSEChannel(
 			this.messagesStore.filterStore.messsagesSSEConfig.type,
 			query,
 			this.onPrevSoftFilterChannelResponse,
 			this.onLoadingError,
+			interval
+				? e => {
+						if (query.startTimestamp - e.timestamp > interval * 1000 * 60) {
```
rzvdv, line 345: we should somehow compare it to our main channel that runs without filters

src/stores/messages/MessagesDataProviderStore.ts
VekesserVladimir: Yes

rzvdv (approved):

PR #118, opened 2021-03-18
TH2 1583. Implement mesage body rule configurator

VitalyMashanov:


YourWeebyNeighbor (changes requested):
I don't like how this hack looks. Let's implement a dedicated component that accepts a single value.

rzvdv (approved):

YourWeebyNeighbor (approved):

PR #123, opened 2021-03-22
Update search form

icefleen:


src/stores/SearchStore.ts
```diff
@@ -45,7 +45,7 @@ export type SearchResult = EventTreeNode | EventMessage;
 
 export type SearchHistory = {
 	timestamp: number;
-	results: Array<SearchResult>;
+	results: Record<string, Array<SearchResult>>;
```
rzvdv, line 48: Some users already have search results saved in localstorage and their will not be able to use app unless they clear local storage themselves

i got
Uncaught TypeError: Cannot read property 'messageId' of undefined SearchPanelResults.tsx:64

src/components/search-panel/SearchPanelForm.tsx
```diff
@@ -134,20 +132,53 @@ const SearchPanelForm = (props: Props) => {
...
+						<div className='search-panel__fields'>
+							<div className='filter-row'>
+								<div className='filter-row__label'>Search for</div>
+								<SearchTypeSwitcher formType={formType} setFormType={setFormType} />
+							</div>
+							<div className='filter-row'>
+								<div className='filter-row__label'>Time Limit</div>
+								<FilterDatetimeInput {...endTimestampInput} />
+							</div>
+							{config.map(rowConfig => (
```
rzvdv, line 161: I thought we move result count limit somewhere. currently it takes whole row, but it doesnt need it. I dont know maybe move it to same row as `Search for` ? @YourWeebyNeighbor what do you think?

src/components/search-panel/SearchPanelResults.tsx
```diff
@@ -49,16 +51,31 @@ const SearchPanelResults = (props: SearchPanelResultsProps) => {
...
 	function computeKey(index: number) {
-		const item = resultGroups[index][0] || 'key';
+		const [, results] = resultGroups[index];
+		const item = results[0];
 
 		return isEventNode(item) ? item.eventId : item.messageId;
 	}
 
+	const onScroll = (e: React.UIEvent<HTMLDivElement, UIEvent>) => {
+		if (e.target === e.currentTarget) {
```
rzvdv, line 66: It's bugged when after collapsing form, results take more space and items take less space then results height, so after onScrollEvent occurs its scrollTop is 0 and its collapsing

src/stores/SearchStore.ts
icefleen: We'll think about it. @YourWeebyNeighbor says we can validate localStorage data and clear it if we got an invalid data format. It'll be resolved in the future

src/components/search-panel/SearchPanelForm.tsx
icefleen: Done
image

src/components/search-panel/SearchPanelResults.tsx
icefleen: I removed this functionality temporarily

YourWeebyNeighbor (approved):

PR #126, opened 2021-03-24
TH2-1670

VitalyMashanov:


src/components/graph/search/GraphSearch.tsx
```diff
@@ -83,8 +86,16 @@ function GraphSearch(props: Props) {
 
 	React.useEffect(() => {
 		const ac = new AbortController();
-
+		const abort = () => {
+			ac.abort();
+		};
 		if (currentValue) {
+			if (currentValue.match(dateRegexp)) {
+				const time = moment(currentValue).utc().valueOf();
```
rzvdv, line 94: if you want to parse date as utc you should do `moment.utc(date)` note `moment(date).utc()` 

and all these should be handled inside GraphSearchInput

PR #127, opened 2021-03-26
Link sharing improvement

VitalyMashanov:


rzvdv:
This doesn't solve the problem but makes some workaround which will confuse user even more. Don't modify sessions. We need to find a way to prevent default behavior of messages store on initial loading

YourWeebyNeighbor:
Seems like a dirty fix. Let's leave it for now and solve it later.

PR #128, opened 2021-03-29
replace-multiple-string-input-submit-key-codes.

VitalyMashanov:


src/components/filter/row/MultipleStringFIlterRow.tsx
```diff
@@ -95,7 +95,7 @@ export default function MultipleStringFilterRow({ config }: MultipleStringFilter
 							key={index}
 							size='small'
 							removeIconType='white'
-							submitKeyCodes={[KeyCodes.SPACE]}
```
rzvdv, line 98: Update placeholder

rzvdv (approved):

PR #178, opened 2021-05-07
messageType overflow in message cards

VitalyMashanov:


rzvdv (changes requested):
There is a problem with flex element with children and text-overflow

rzvdv (approved):

PR #195, opened 2021-05-26
filters-sharing.

VitalyMashanov:


src/stores/FiltersHistoryStore.ts
```diff
@@ -174,9 +166,38 @@ class FiltersHistoryStore {
...
 			FiltersHistoryType<EventFilterState | MessageFilterState>
 		>(IndexedDbStores.FILTERS_HISTORY);
-
 		this.filterHistory = history.sort(sortByTimestamp);
+		this.initialized = true;
+	};
+
+	@action
+	private addEventHistoryItem = async (newItem: FiltersHistoryType<EventFilterState>) => {
+		if (this.eventsHistory.some(({ filters }) => isEqual(filters, newItem.filters))) return;
```
rzvdv, line 175: I think it's better to change this behaviour.
If we find same filter lets update it's timestamp.

rzvdv (changes requested):
We should notify user if shared filter was successfully saved

src/stores/FiltersHistoryStore.ts
```diff
@@ -113,43 +115,33 @@ class FiltersHistoryStore {
...
-			filters: equilizedFilter,
-			isPinned: false,
-		};
-
-		this.indexedDb.addDbStoreItem(IndexedDbStores.FILTERS_HISTORY, filter);
-		this.filterHistory = [...this.filterHistory, filter];
+		if (this.initialized) {
+			this.addMessageHistoryItem(filter);
+		} else {
+			this.init().then(() => {
```
rzvdv, line 141: we call this function twice now.
let's make initialized observable property
then we can use `when` from mobx
`await when(() => this.initialized)`
then add

src/components/filters-history/FiltersHistoryItem.tsx
```diff
@@ -146,16 +161,25 @@ const FiltersHistoryItem = (props: Props) => {
...
 				{moment.utc(item.timestamp).format(FILTER_HISTORY_DATE_FORMAT)}
-				<button
-					className={pinButtonClassname}
-					onClick={onFilterPin}
-					ref={pinButtonRef}
-					title={item.isPinned ? 'Unpin filter' : 'Pin filter'}>
-					<i></i>
-				</button>
-			</p>
+				<div className='controls'>
```
rzvdv, line 166: .filter-history-item__controls

src/stores/FiltersHistoryStore.ts
```diff
@@ -113,43 +117,35 @@ class FiltersHistoryStore {
...
-			type,
-			filters: equilizedFilter,
-			isPinned: false,
-		};
-
-		this.indexedDb.addDbStoreItem(IndexedDbStores.FILTERS_HISTORY, filter);
-		this.filterHistory = [...this.filterHistory, filter];
+	public onEventFilterSubmit = async (
+		newItem: FiltersHistoryType<EventFilterState>,
+		notify?: boolean,
```
rzvdv, line 122: delete notify
this function returns promise so we can chain with `then` or await where this function is being called

src/stores/FiltersHistoryStore.ts
```diff
@@ -174,9 +170,58 @@ class FiltersHistoryStore {
...
+		const existedFilter = this.eventsHistory.find(({ filters }) =>
+			isEqual(filters, newItem.filters),
+		);
+		if (existedFilter) {
+			existedFilter.timestamp = newItem.timestamp;
+			this.indexedDb.updateDbStoreItem(IndexedDbStores.FILTERS_HISTORY, {
+				...toJS(observable(existedFilter)),
+				timestamp: newItem.timestamp,
+			});
+			return;
```
rzvdv, line 188: if there is already an if block use else,delete return

src/stores/FiltersHistoryStore.ts
```diff
@@ -174,9 +170,58 @@ class FiltersHistoryStore {
...
+		this.initialized = true;
+	};
+
+	@action
+	private addEventHistoryItem = async (newItem: FiltersHistoryType<EventFilterState>) => {
+		const existedFilter = this.eventsHistory.find(({ filters }) =>
+			isEqual(filters, newItem.filters),
+		);
+		if (existedFilter) {
+			existedFilter.timestamp = newItem.timestamp;
```
rzvdv, line 183: You can update timestamp since you chose timestamp to be key in indexeddb. this will create new filter

src/stores/FiltersHistoryStore.ts
```diff
@@ -113,43 +117,35 @@ class FiltersHistoryStore {
...
+		notify?: boolean,
+	) => {
+		if (isEmptyFilter(newItem.filters)) return;
+
+		const filter = getEquilizedItem(newItem);
+
+		await when(() => this.initialized);
+		this.addMessageHistoryItem(filter);
+		if (notify) {
+			showNotification('Filter successfully saved');
```
rzvdv, line 147: Use toasts which appear in Notifier instead

src/stores/FiltersHistoryStore.ts
```diff
@@ -170,13 +155,62 @@ class FiltersHistoryStore {
...
+		this.initialized = true;
+	};
+
+	@action
+	private addEventHistoryItem = async (newItem: FiltersHistoryType<EventFilterState>) => {
+		const existedFilter = this.eventsHistory.find(({ filters }) =>
+			isEqual(filters, newItem.filters),
+		);
+		if (existedFilter) {
+			this.indexedDb.deleteDbStoreItem(IndexedDbStores.FILTERS_HISTORY, existedFilter.timestamp);
```
rzvdv, line 182: delete from filtersHistory aswell

src/stores/FiltersHistoryStore.ts
```diff
@@ -170,13 +155,62 @@ class FiltersHistoryStore {
 		}
 	};
 
+	@action
+	public showSuccessNotification = () => {
+		this.notificationsStore.addMessage({
+			type: 'success',
+			notificationType: 'success',
+			description: 'Filter successfully saved!',
```
rzvdv, line 163: let's change that to `Event/Message filter successfully saved`

rzvdv (changes requested):
update version

rzvdv (approved):

PR #214, opened 2021-07-07
Date input focus

Ryzhkov-Alexey:
focusing the date input field by key combination shift + t


src/components/graph/search/GraphSearchInput.tsx
```diff
@@ -281,15 +283,44 @@ function GraphSearchInput(props: Props) {
 		}
 	}
 
+	function preventDefaultHandlers(handlers: Object) {
+		const newHandlers: any = {};
```
rzvdv, line 287: let's not use any
instead we can use
```
	type KeyboardHandler = {
		[key: string]: (keyEvent?: KeyboardEvent | undefined) => void;
	};
```

and make handlers type `KeyboardHandler`
and `const newHandlers: KeyboardHandler`

Ryzhkov-Alexey: I will fix it, thanks

rzvdv (changes requested):
Delete package-lock.json changes. I'm not sure how they have come up. You should have installed dependencies by cd into the project and running npm install

rzvdv (approved):

PR #215, opened 2021-07-08
New icon

Ryzhkov-Alexey:
new icons


rzvdv (changes requested):
There are too many changes in package-json.lock, could you delete that file from commit ?

src/styles/messages.scss
```diff
@@ -166,7 +166,7 @@
 		&.json {
 			position: relative;
 			&:before {
-				content: '{~}';
+				content: url(../../resources/icons/json.svg);
```
rzvdv, line 169: json icon aligning seems to be a litte off

Ryzhkov-Alexey: I will fix it, thanks

rzvdv (changes requested):
Same icons are used in RuleEditor component (settings icon on top right). We need to change them aswell and delete old icons

rzvdv (approved):

PR #217, opened 2021-07-13
[TH2-2082] Fix messages filter session filter.

artem-davidovich:


src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -199,12 +200,28 @@ const MessagesFilterPanel = () => {
 		];
 	}, [messagesStore.messageSessions, sessionsStore.sessions]);
 
+	const updateStreams = (newSteams: string[]) => {
+		if (isFocused) {
+			setStreams(newSteams);
+		} else {
+			searchStore.stopSearch();
```
rzvdv, line 207: SearchStore is for SearchPanel, i don't get why we should stop search on SearchPanel when sessions on message tab are changed.
I don't understand why we need isFocused state aswell.

Since we refetch messages on sessions change, submit button near sessions input is not longer needed

rzvdv (changes requested):

src/components/filter/MessagesFilterPanel.tsx
artem-davidovich: > I don't understand why we need isFocused state aswell.
Filters should be applied after onblur event, that is why I decided use isFocused state.

src/components/filter/row/MultipleStringFIlterRow.tsx
```diff
@@ -28,14 +28,25 @@ import {
 
 interface MultipleStringFilterRowProps {
 	config: FilterRowMultipleStringsConfig;
+	setIsInputFocused?: (value: Boolean) => void;
```
rzvdv, line 31: use primitive types

rzvdv (changes requested):
it doesnt work when i select session from autocomplete and click outside

if we want to set filterStore.filter on focusing out we need to do it inside useEffect. The problem is that when you submit value or click on another Bubble, AutocompleteInput unmounts and loses focus, then another AutocompleteInput mounts. State isFocused will quikly switch between true and false. and useEffect function would fire unnecessary.

AutocompleteInput is a mess right now. is see it fires `onBlur()` inside of onClickOutside which is not right i believe. delete that.

Then what we can do is
 when input is focused we immediately set `isFocused: true`, but when it looses focus it should have some updater function with a delay for example 100ms. If is focused changes to true this updater function should be cancelled

artem-davidovich:
@rzvdv 
> AutocompleteInput is a mess right now. is see it fires onBlur() inside of onClickOutside which is not right i believe. delete that. and pass onBlur function to inputProps inside AutocompleteInput

If we pass onBlur function to inputProps inside AutocompleteInput, focus will be loosed when a bubble selected. That is why, I think it is correct, when onClickOutside fires onBlur event.

PR #219, opened 2021-07-20
Embedded mode

Ryzhkov-Alexey:


icefleen (changes requested):

src/stores/RootStore.ts
```diff
@@ -182,7 +188,16 @@ export default class RootStore {
...
+		const isEmbedded = searchParams.get('viewMode') === 'embedded';
+		if (isEmbedded) {
+			const eventId = searchParams.get('eventId');
+			const messageId = searchParams.get('messageId');
+			if ((eventId !== null && messageId === null) || (eventId === null && messageId !== null)) {
+				return isEmbedded;
+			}
+			throw new Error('message Id and event Id cannot be specified at the same time');
+		}
+		return isEmbedded;
```
icefleen, line 200:
```
const isEmbedded = searchParams.get('viewMode') === 'embedded';

if (isEmbedded) {
	const eventId = searchParams.get('eventId');
	const messageId = searchParams.get('messageId');
	
	if (eventId && messageId) {
		throw new Error('message Id and event Id cannot be specified at the same time');
	} else if (!eventId && !messageId) {
		// Error again
	}
}

return isEmbedded;
```

src/stores/RootStore.ts
```diff
@@ -123,6 +123,12 @@ export default class RootStore {
 
 	private parseUrlState = (): WorkspacesUrlState | null => {
 		try {
+			if (
+				(!this.isEmbedded && window.location.search.split('&').length > 1) ||
+				(this.isEmbedded && window.location.search.split('&').length > 2)
```
icefleen, line 128: you can delete second condition due to change below

src/styles/root.scss
```diff
@@ -98,6 +98,10 @@ input {
 		}
 	}
 
+	&__embedded {
+		background-color: #fff;
```
icefleen, line 102: create new color variable in `src/styles/common/_colors.scss` for embeeded layout and use it

icefleen (changes requested):

src/components/event/EventDetailInfoCard.tsx
```diff
@@ -23,20 +23,20 @@ import { getEventStatus } from '../../helpers/event';
...
 import { EventAction, EventTreeNode } from '../../models/EventAction';
-import { useRootStore, useSelectedStore } from '../../hooks';
+import { useSelectedStore } from '../../hooks';
 
 interface Props {
 	node: EventTreeNode;
 	event: EventAction | null;
 	childrenCount?: number;
 	children?: React.ReactNode;
+	isEmbedded: boolean;
```
icefleen, line 33: make this field optional (use `?` symbol as in case of children and childrenCount above), then remove all `isEmbedded={false}` in other places

src/components/embedded/EmbeddedWorkspace.tsx
```diff
@@ -10,6 +10,7 @@ function EmbeddedWorkspace({ workspace }: { workspace: WorkspaceStore }) {
 			<EventDetailInfoCard
 				node={workspace.eventsStore.selectedNode}
 				event={workspace.eventsStore.selectedEvent}
+				isEmbedded={true}
```
icefleen, line 13: you can just write `isEmbedded` without `={true}`

src/components/message/message-card/MessageCard.tsx
```diff
@@ -41,6 +40,7 @@ const HUE_SEGMENTS_COUNT = 36;
 
 export interface OwnProps {
 	message: EventMessage;
+	isEmbedded: boolean;
```
icefleen, line 43: same, the field must be optional

icefleen (changes requested):

icefleen (dismissed):

rzvdv (changes requested):
1) App has layout `grid-template-rows: 100px 1fr`
if message is small it's ok if its a big message it will not fully fit in 100px.
2) Need to add some loading indicators, right now it's a blank screen until message is loaded
3) target message should be loaded separately, right now app still loads previous/next messages via SSE
4) Messages highlights when after it has been loaded. which is unnecessary. We made it to highlight target message inside list

src/components/embedded/EmbeddedApp.tsx
```diff
@@ -0,0 +1,24 @@
...
+	const messageId = searchParams.get('messageId');
+	window.history.replaceState({}, '', window.location.pathname);
+
+	if (eventId) {
+		return <EmbeddedEvent eventId={eventId} />;
+	}
+	if (messageId) {
+		return <EmbeddedMessage messageId={messageId} />;
+	}
+	return <p>Message or event ID not found</p>;
```
rzvdv, line 21: if we get here it means there is no id provided. So lets throw that error

src/components/embedded/EmbeddedApp.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export default observer(EmbeddedApp);
```
rzvdv, line 24: delete observer

src/components/embedded/EmbeddedEvent.tsx
```diff
@@ -0,0 +1,40 @@
...
+import SplashScreen from '../SplashScreen';
+
+function EmbeddedEvent({ eventId }: { eventId: string }) {
+	const [event, setEvent] = useState<EventAction | null>(null);
+
+	useEffect(() => {
+		getMessage();
+	}, []);
+
+	async function getMessage() {
```
rzvdv, line 14: function getEvent()

there is a problem here that on error component is stuck on showing loading.
part of the problem is that api.events.getEvent return null if event hasnt been found.
i think we made api return null if there is an error because we handled all errors inside fetch-intercept.ts.

What we can do is to call `fetch(/th2-commonv3/backend/event/${id}?${params})` in this component as well as inside EmbeddedMessage. then handle error and throw it.

we can take this logic from fetch-intercept which returns error message based on status and make another function getErrorMessage to use it here, in EmbeddedMessage component, and in fetch-intercept.ts

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+	const [viewType, setViewType] = useState(MessageViewType.JSON);
+
+	useEffect(() => {
+		getMessage();
+	}, []);
+
+	async function getMessage() {
+		const res = await api.messages.getMessage(messageId);
+		setMessage(res);
+		setViewType(message?.viewType ? message.viewType : MessageViewType.JSON);
```
rzvdv, line 18: actually viewType doesn't exist in message i think. we should delete that property from interface

rzvdv (changes requested):

src/components/message/message-card/raw/MessageRaw.tsx
```diff
@@ -19,14 +19,51 @@ import { observer } from 'mobx-react-lite';
...
 interface Props {
 	rawContent: string;
 	messageId: string;
 	renderInfo: () => React.ReactNode;
+	isEmbedded?: boolean;
+	viewType: MessageViewType.ASCII | MessageViewType.BINARY;
 }
 
-function MessageRaw({ rawContent, messageId, renderInfo }: Props) {
+interface EmbeddedProps {
```
rzvdv, line 32: we can get that prop `isDetailed` in MessageCard and pass down so there would be no need to create a separate component

src/components/message/message-card/MessageCard.tsx
```diff
@@ -47,11 +47,173 @@ export interface RecoveredProps {
...
+	isSoftFiltered?: boolean;
+	isContentBeautified?: boolean;
+	toogleMessagePin?: () => void;
+	isEmbedded?: boolean;
+}
+
 interface Props extends OwnProps, RecoveredProps {}
 
-function MessageCardBase({ message, viewType, setViewType }: Props) {
+export function MinimalMessageCard({
```
rzvdv, line 64: let's refactor MessageCardBase into arrow function and wrap it in observer. 
then delete `const MessageCard = observer(MessageCardBase);`
then rename MessageCardBase to MessageCard and MinimalMessageCard to MessageCardBase.

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+
+	if (errorStatus) {
+		throw new Error(errorStatus);
+	}
+
+	if (message) {
+		return (
+			<MessageCardBase isEmbedded message={message} setViewType={setViewType} viewType={viewType} />
```
rzvdv, line 30: handle overflow, if messages is bigger than viewport its not scrollable

rzvdv:
MessageBodyCard is using stores and throws errors
For some reason there is no header in message when it's json mode

src/styles/embedded.scss
```diff
@@ -0,0 +1,4 @@
+.embedded-message-wrapper {
```
rzvdv, line 1: events should have overflow handled aswell

src/components/embedded/EmbeddedApp.tsx
```diff
@@ -0,0 +1,23 @@
...
+	const messageId = searchParams.get('messageId');
+	window.history.replaceState({}, '', window.location.pathname);
+
+	if (eventId) {
+		return <EmbeddedEvent eventId={eventId} />;
+	}
+	if (messageId) {
+		return <EmbeddedMessage messageId={messageId} />;
+	}
+	throw new Error('Message or event ID not found');
```
rzvdv, line 20: Lets change that message to Provide messageId or eventId

src/styles/embedded.scss
```diff
@@ -0,0 +1,4 @@
...
+    height: 100%;
```
rzvdv, line 2: add license like in other scss files

rzvdv (approved):

PR #220, opened 2021-07-20
Dragable replay

Ryzhkov-Alexey:


rzvdv (changes requested):
this modal is centered using this css
```
  position: absolute;
  top: 50%;
  left: 50%;
  transform: translate(-50%, -50%);
```

Framer motion uses transform on drag to change div position. So when you drag it resets initial transform we had `translate(-50%, -50%)` and applies its own transform styles causing it to jump on initial drag.
Now we need to center modal without using transform. E.g for left we can use left: calc(50% - 400px), for top we can use something similar

src/styles/messages.scss
```diff
@@ -801,15 +801,15 @@
 	display: grid;
 	row-gap: 5px;
 	position: absolute;
-	top: 50%;
-	left: 50%;
-	transform: translate(-50%, -50%);
+	top: 35%;
```
rzvdv, line 804:
```
top: calc(50% - 200px);
left: calc(50% - 400px);
```

rzvdv (changes requested):
Currently  you can click at any point inside modal and it becomes draggable, that's not exactly how we want it to work. right now if you try to select text anywhere modal moves. We need to add some sort of dragging zone, it could be top area of the modal for example, also when hovering it should change cursor to 'move'

src/components/message/MessageReplayModal.tsx
```diff
@@ -49,6 +50,8 @@ function MessageReplayModal() {
 
 	const [isOpen, setIsOpen] = React.useState(false);
 
+	const [draggable, setDragable] = React.useState(true);
```
rzvdv, line 53: this doesn't work when you drag vertically.

you probably need this https://www.framer.com/docs/gestures/###dragcontrols

rzvdv (approved):

PR #221, opened 2021-07-20
Opening message by link

Ryzhkov-Alexey:


rzvdv (changes requested):
Never change state inside component render, this will be called each time component is rendered.  if you want to change state inside component probably you should do it inside useEffect hook

If you need to make messages expand if messageId is present you should do in inside parseUrl state in RootStore
and inside return statement you should change layout to layout: messageId ? [0, 100] : [100, 0],

rzvdv (approved):

rzvdv (approved):

PR #222, opened 2021-07-23
Timestamp position selection

Ryzhkov-Alexey:


src/components/graph/search/GraphSearch.tsx
```diff
@@ -257,13 +266,7 @@ function GraphSearch(props: Props) {
...
 					{mode === 'confirm' && refreshPanels && (
 						<GraphSearchConfirm
-							onConfirm={() => {
-								if (timestamp && refreshPanels) {
-									refreshPanels(timestamp);
-								}
-								setShowModal(false);
-								setMode('timestamp');
-							}}
+							onConfirm={(positon: TimestampPosition) => confirm(positon)}
```
rzvdv, line 269: `onConfirm={confirm}`

rzvdv (changes requested):
there is a bug that submits timestamp and changes graph time range before GraphSearchConfirm counts to 0, let's fix that,
we also want to center timeline properly after submitting

PR #233, opened 2021-08-28
TH2-2183. Add hints to explain Search - Parent Event field work

VitalyMashanov:
* hints for all fields added


src/components/search-panel/SearchPanelForm.tsx
```diff
@@ -104,6 +104,7 @@ const SearchPanelForm = () => {
 		setValue: getFormStateUpdater('parentEvent'),
 		type: 'event-resolver',
 		id: 'parent-event',
+		placeholder: 'matches parent event children if they are exist',
```
rzvdv, line 107: 'matches events by specified parent event id' is better i think

rzvdv (approved):

rzvdv (approved):

PR #235, opened 2021-08-31
TH2-1868. Bookmarks' scrollbar doesn't appear when there are one or two extra bookmarks in list

VitalyMashanov:
* fixed

### DIFF
--- a/src/styles/bookmarks.scss
+++ b/src/styles/bookmarks.scss
@@ -19,13 +19,11 @@
 .bookmarks-panel {
 	height: 100%;
 	background-color: #fff;
-	display: flex;
-	flex-direction: column;
+	display: grid;
+	grid-template-rows: auto 1fr;
 	
 	&__container {
-		padding: 8px 2px 8px 8px;
-		width: inherit;
-		height: inherit;
+		padding: 8px 2px 0 8px;
 	}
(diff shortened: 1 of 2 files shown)


src/styles/bookmarks.scss
```diff
@@ -17,23 +17,19 @@
...
 	background-color: #fff;
-
+	display: flex;
+	flex-direction: column;
+	
 	&__container {
-		padding: 8px;
-		width: 100%;
-		height: 100%;
+		padding: 8px 2px 8px 8px;
```
rzvdv, line 26: lets remove bottom padding

src/styles/bookmarks.scss
```diff
@@ -17,23 +17,19 @@
 @import './common/vars';
 
 .bookmarks-panel {
-	width: 100%;
 	height: 100%;
 	background-color: #fff;
-
+	display: flex;
```
rzvdv, line 22: i think grid is better for this kind of things
```
display: grid;
grid-template-rows: auto 1fr;
```
and then delete
```
		width: inherit;
		height: inherit;
```

rzvdv (approved):

rzvdv (approved):

PR #238, opened 2021-09-03
TH2-2035 Don't reset message filter when user chooses another event.

VitalyMashanov:
* fixed


src/stores/messages/MessagesStore.ts
```diff
@@ -317,22 +317,21 @@ export default class MessagesStore {
 		filter change hint window is shown to a user. And it is up to him to decide
 		if he wants to reset streams to message(s) streams and update filters
 	*/
-	private handleFilterHint = (message: EventMessage | EventMessage[]): boolean => {
+	private handleFilterHint = async (message: EventMessage | EventMessage[]): Promise<boolean> => {
```
rzvdv, line 320: this function is now async, but is it waited where it is used?

src/stores/messages/MessagesFilterStore.ts
```diff
@@ -139,10 +139,8 @@ export default class MessagesFilterStore {
 
 	@action
 	public resetMessagesFilter = (initFilter: Partial<MessagesFilter> = {}) => {
-		const filter = getDefaultMessagesFiltersState(this.searchStore.messagesFilterInfo);
 		const defaultMessagesFilter = getDefaultMessagesFilter();
```
rzvdv, line 142: why we deleted that?

src/stores/messages/MessagesStore.ts
```diff
@@ -322,22 +324,21 @@ export default class MessagesStore {
...
-		const sseFilter = this.filterStore.sseMessagesFilter;
-		const areFiltersApplied = [
-			sseFilter
-				? [sseFilter.attachedEventIds.values, sseFilter.body.values, sseFilter.type.values].flat()
-				: [],
-		].some(filterValues => filterValues.length > 0);
+		const matchMessageParams = this.filterStore.filterParams;
+
+		const hintMessagesMatch = await Promise.all(
+			this.hintMessages.map(hm => this.api.messages.matchMessage(hm.messageId, matchMessageParams)),
```
rzvdv, line 338: filtering is async, so we need to show user that something is loading

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -293,7 +293,11 @@ const MessagesFilterPanel = () => {
 				renderFooter={renderFooter}
 			/>
 			<MessageReplayModal />
-			<MessageFilterWarning />
+			{messagesStore.checkingAttachedMessages ? (
```
rzvdv, line 296: There is already loader in this component. 2 loaders would be to much i think

FilterPanel accepts isLoading prop, we can pass it there

src/stores/messages/MessagesStore.ts
```diff
@@ -80,6 +80,9 @@ export default class MessagesStore {
 	@observable
 	public showFilterChangeHint = false;
 
+	@observable
+	public checkingAttachedMessages = false;
```
rzvdv, line 84: can we rename it to filteringAttachedMessages ?

src/stores/messages/MessagesStore.ts
```diff
@@ -322,22 +332,26 @@ export default class MessagesStore {
...
-				? [sseFilter.attachedEventIds.values, sseFilter.body.values, sseFilter.type.values].flat()
-				: [],
-		].some(filterValues => filterValues.length > 0);
+		this.setCheckingAttachedMessages(true);
+
+		const matchMessageParams = this.filterStore.filterParams;
+
+		const hintMessagesMatch = await Promise.all(
+			this.hintMessages.map(hm => this.api.messages.matchMessage(hm.messageId, matchMessageParams)),
+		).then(match => {
```
rzvdv, line 349: what if request will fail for some reason?

let's replace `then` with `finally`

rzvdv (approved):

PR #240, opened 2021-09-08
TH2-1899. When using 'Get Workspace Link', attached message to selected event is not displayed in new tab

VitalyMashanov:
*fixed


rzvdv:
I think this is a bug or report-data-provider, not viewer

PR #242, opened 2021-09-09
TH2-2313.  Critical error "Cannot read property 'negative' of undefined" after opening a saved report

VitalyMashanov:
* fixed


rzvdv (changes requested):
I believe error is in the SearchPanel. when we have old saved search results with old filters, and we trying to extract filter values by current filter info

rzvdv (dismissed):

rzvdv (approved):

PR #243, opened 2021-09-09
Fix message search

TheodorSchmidt:
Invalid messages are marked as invalid. Search is highlighted in red only if invalid messages are entered


src/components/search-panel/SearchPanelForm.tsx
```diff
@@ -73,6 +73,21 @@ const SearchPanelForm = () => {
 		];
 	}, [messageSessions, sessionsStore.sessions]);
 
+	const invalidCheck: boolean = React.useMemo(() => {
```
rzvdv, line 76: let's rename it to areSessionsInvalid.

You can also omit type annotation, typescript will infer it from the value you return.
and validation can be simplified
```
form.stream.length > 0 && form.stream.some(stream => !messageSessions.includes(stream.trim()))
```

rzvdv (approved):

PR #244, opened 2021-09-09
Th2-2288. datetime input - find nearest event automatically

molotgor:


src/components/graph/search/GraphSearch.tsx
```diff
@@ -75,6 +75,8 @@ function GraphSearch(props: Props) {
 	const wrapperRef = React.useRef<HTMLDivElement>(null);
 	const modalRef = React.useRef<HTMLDivElement>(null);
 
+	const [timeMode, setTimeMode] = React.useState(false);
```
rzvdv, line 78: i think the better pattern to fetch event is to have submittedTimestamp: Number | null
then in GraphSearchDialog you can have effect

```
	React.useEffect(() => {
		const searchChannels = {
			previous: submittedTimestamp ? getEventSource(SearchDirection.Previous) : null,
			next: submittedTimestamp ? getEventSource(SearchDirection.Next) : null,
		};

		return () => {
			searchChannels.next?.stop();
			searchChannels.previous?.stop();
		};
	}, [submittedTimestamp]);

```
  in your current implementation you are searching not closest event, but first that you received, this might be a problem.

Also after event has been found this component makes 4 another http requests with ?probe=true for some reason

molotgor: Http request were sent because of fetchObjectById. There were no check for time and because of that were sent requests for found id and for inputted time.

src/components/graph/search/GraphSearch.tsx
```diff
@@ -164,23 +151,23 @@ function GraphSearch(props: Props) {
 	);
 
 	const onModeSelect = (newMode: GraphSearchMode) => {
+		setTimestamp(null);
```
icefleen, line 154: why you're resetting the timestamp?

src/components/graph/search/GraphSearch.tsx
```diff
@@ -252,39 +240,25 @@ function GraphSearch(props: Props) {
...
-							<>
-								<button className={dateButtonClassName} onClick={() => onModeSelect('timestamp')}>
-									Timestamp
-								</button>
-								<button className={idButtonClassName} onClick={() => onModeSelect('history')}>
-									History
-								</button>
-							</>
-						)}
+						<>
```
icefleen, line 248: since the condition has been removed, now you can delete this React.Fragment (`<></>`)

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -74,12 +82,17 @@ const GraphSearchDialog = (props: Props) => {
 
 	const filteredSearchHistory = React.useMemo(() => {
 		if (!value) return sortedSearchHistory;
-		return sortedSearchHistory.filter(
+		const idIndex = sortedSearchHistory.map(historyItem => getItemId(historyItem.item));
```
icefleen, line 85: pls use a more informative name for the `idIndex` variable

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -139,14 +152,62 @@ const GraphSearchDialog = (props: Props) => {
...
+			searchChannels.previous?.close();
+		};
+
+		const onChannelResponse = (ev: any) => {
+			stopSearch();
+			const data = (ev as MessageEvent).data;
+			const parsedEvent = JSON.parse(data);
+			const id = getItemId(parsedEvent);
+			ac.current = new AbortController();
+			onValueChange(id, ac.current);
```
icefleen, line 197: i think you need to call the `onSearchSuccess` function, because `onValueChange` calls one more api request to find an event by id, while you already have this event in the `parsedEvent`

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -195,6 +256,11 @@ const GraphSearchDialog = (props: Props) => {
 
 	const fetchObjectById = (id: string, abortController: AbortController) => {
 		const searchTimestamp = moment.utc().valueOf();
+		if (
```
icefleen, line 259: if you replace `onValueChange` to `onSearchSuccess` (see above), you can remove this condition

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -28,6 +28,10 @@ import { useDebouncedCallback, useRootStore } from '../../../hooks';
 import KeyCodes from '../../../util/KeyCodes';
 import { indexedDbLimits, IndexedDbStores } from '../../../api/indexedDb';
 import { GraphSearchResult } from './GraphSearch';
+import { SearchDirection } from '../../../models/search/SearchDirection';
+import { DateTimeMask } from '../../../models/filter/FilterInputs';
+
+type SSESearchDirection = SearchDirection.Next | SearchDirection.Previous;
```
icefleen, line 34: unused code fragment

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -139,14 +152,62 @@ const GraphSearchDialog = (props: Props) => {
...
 				onValueChange(id, ac.current);
 			}
 		}
 	}, [submittedId]);
 
+	React.useEffect(() => {
+		const searchChannels = {
+			previous: submittedTimestamp
+				? api.sse.getEventSource({
+						type: 'event',
```
icefleen, line 166: are you sure that we don't need to search messages?

icefleen (changes requested):

src/components/graph/search/GraphSearchDialog.tsx
molotgor: yes, task says that i need to find nearest event

src/components/graph/search/GraphSearch.tsx
molotgor: That's made to prevent search happening when mode changes from timestamp to history by switcher rather than by submit.

src/components/graph/search/GraphSearchDialog.tsx
molotgor: Even with change this condition prevents unneeded requests even with change to onValueChange and it prevents requests when in history mode inputted date

icefleen: ok. can we move this condition to `onValueChange` before calling `fetchObjectById`? if yes, let's do it. and create new function for this logic like `isTimestamp(value: string): boolean` in `src/helpers/date.ts` and use it.

icefleen (dismissed):
looks good to me. i'll ask Anatoly about merging

rzvdv (approved):

rzvdv (approved):

PR #245, opened 2021-09-13
[TH2-2312] Make correct messages rendering with new api

osinkinroman:


src/models/MessageBody.ts
```diff
@@ -14,18 +14,27 @@
  * limitations under the License.
  ***************************************************************************** */
 
-export default interface MessageBody {
-	metadata: {
-		id: {
-			connectionId: {
-				sessionAlias: string;
+export default interface MessageBody extends Array<MessageBody> {
```
rzvdv, line 17: let's make it similar to events body.
we will have MessageBodyPayload type
and `type MessageBody = MessageBodyPayload[]`;
then instead of passing index for rendering some payload you will pass this payload,

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -43,13 +43,18 @@ function EmbeddedMessage({ messageId }: { messageId: string }) {
...
 		return (
-			<div className='embedded-wrapper'>
-				<MessageCardBase
-					isEmbedded
-					message={message}
-					setViewType={setViewType}
-					viewType={viewType}
-				/>
+			<div>
+				{message.body?.map((item, intemIndex) => (
```
rzvdv, line 47: There should be a component which renders message body instead of manually iterating over its body

rzvdv (changes requested):

src/models/MessageBody.ts
```diff
@@ -14,18 +14,27 @@
...
+	subsequenceId: [number];
```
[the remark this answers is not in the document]
osinkinroman: image
In new api this field looks like [1], so I literally did as it is here, maybe it is really wrong, I don't know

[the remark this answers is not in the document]
osinkinroman: Выдел1ение_059
Here is example of new api, so subsequenceId is [1], so I decided to do literally like here, however I don't know is it right to do like that

rzvdv:
I think RecoverableMessageCard should now wrap not MessageCard component, but MessageCardBase.
and you will use stateKey like so
`keyForMessage(`${props.message.messageId}-${subsequenceId}`)`

Also move useEffect with [viewType] from MessageCard to MessageCardBase, and there you should change messageId to `messageId-subsequenceId`

webpack/webpack.dev.js
```diff
@@ -14,65 +14,65 @@
  * limitations under the License.
  ******************************************************************************/
 
-const webpackMerge = require('webpack-merge');
-const commonConfig = require('./webpack.common');
-const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
-const { appSrc } = require('./paths');
+ const webpackMerge = require('webpack-merge');
```
rzvdv, line 17: this file shouldnt be in pr

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "th2-rpt-viewer",
-  "version": "3.1.74",
```
rzvdv, line 3: if you want to update version do it in package.json not in package-lock.json

src/models/MessageBody.ts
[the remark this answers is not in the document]
rzvdv: change it to `number[]`

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -38,18 +39,23 @@ function EmbeddedMessage({ messageId }: { messageId: string }) {
 	}
 
 	if (errorStatus) {
-		throw new Error(errorStatus);
+		throw new Error();
```
rzvdv, line 42: We need error message

src/components/message/message-card/MessageCard.tsx
```diff
@@ -39,157 +34,168 @@ export interface RecoveredProps {
...
-			isAttached={isAttached}
-			isContentBeautified={isContentBeautified}
-			isSoftFiltered={isSoftFiltered}
-			toogleMessagePin={toogleMessagePin}
-			isDetailed={isDetailed}
-			sortOrderItems={sortOrderItems}
-		/>
-	);
-});
+// interface Props extends OwnProps, RecoveredProps {}
```
rzvdv, line 37: by commenting this component you disabled highlighting and soft filtering.

we need this component, move this `message.body.map` code inside this component

src/components/embedded/EmbeddedMessageCardBase.tsx
```diff
@@ -0,0 +1,228 @@
...
+	toogleMessagePin?: () => void;
+	isDetailed?: boolean;
+	isEmbedded?: boolean;
+	sortOrderItems?: string[];
+	viewType: MessageViewType;
+	bodyItem: MessageBodyPayload;
+	setViewType: (viewType: MessageViewType) => void;
+}
+
+export function EmbeddedMessageCardBase({
```
rzvdv, line 50: we should reuse MessageCardBase, delete this file

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -38,18 +39,23 @@ function EmbeddedMessage({ messageId }: { messageId: string }) {
...
 		return (
-			<div className='embedded-wrapper'>
-				<MessageCardBase
-					isEmbedded
-					message={message}
-					setViewType={setViewType}
-					viewType={viewType}
-				/>
+			<div>
+				{message.body?.map((item: MessageBodyPayload) => (
```
rzvdv, line 48: move this mapping logic to MessageCard

src/components/message/message-card/MessageCard.tsx
```diff
@@ -127,30 +99,22 @@ const MessageCard = observer(({ message, viewType, setViewType }: Props) => {
...
-
 	function toogleMessagePin() {
 		selectedStore.toggleMessagePin(message);
 	}
 
 	const isExported = messagesStore.exportStore.isExported(message);
 
 	return (
-		<MessageCardBase
+		<RecoverableMessageCard
```
rzvdv, line 109: I thought we will do it like so

```
return (
  <div>
  {props.message.body.map(item => <RecoverableMessageCard item={item}/>)}
  </div>
)
```

src/components/message/message-card-list/MessagesCardList.tsx
```diff
@@ -31,8 +31,33 @@ function MessageCardList() {
 	const messagesStore = useMessagesWorkspaceStore();
 	const messagesDataStore = useMessagesDataStore();
 
+	const [isHighlighted, setHighlighted] = React.useState(false);
+
+	const highlightTimer = React.useRef<NodeJS.Timeout>();
+
+	React.useEffect(() => {
```
rzvdv, line 38: return highlight to MessageCard component, you will have root div element there

PR #249, opened 2021-09-23
TH2-2368 Show latest events button

VekesserVladimir:


rzvdv (changes requested):
I believe EventStore.onRangeChange expects the center timestamp of interval. so when you select current time it will be now +-1/2 interval. so 1/2 of interval will be in the future. I think range should end at current time.

rzvdv (approved):

PR #250, opened 2021-09-23
TH2-2366. Add user preferences page

VitalyMashanov:


src/hooks/useTheme.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useLayoutEffect } from 'react';
+
+export interface Theme {
+	[prop: string]: string;
+}
+
+export default function useTheme(theme: Theme) {
+	useLayoutEffect((): void => {
+		Object.entries(theme).forEach(([variable, value]) => {
+			document.documentElement.style.setProperty(`--${variable}`, value);
```
rzvdv, line 26: i think css variables should be defined on body element

src/components/messages-view-configurator/MessagesViewConfigurator.tsx
```diff
@@ -14,84 +14,110 @@
...
+}
 
-	useOutsideClickListener(modalRef, (e: MouseEvent) => {
-		const isFromAutocomplete = Boolean((e.target as HTMLElement).closest('.rules-autocomplete'));
-		const isFromSelect = Boolean((e.target as HTMLElement).closest('.rules-select-options-list'));
-		if (e.target !== buttonRef.current && !isFromAutocomplete && !isFromSelect) {
-			setIsOpen(false);
-		}
-	});
+const themes: { [k: string]: Theme } = {
```
rzvdv, line 42: not sure why would we manage theme css since we use scss in this project.

i think it's better just to switch classes on body element, each class would have it's own set of css-variables

PR #253, opened 2021-09-29
TH2-1352 Search by parent event name

VekesserVladimir:


src/stores/SearchStore.ts
```diff
@@ -129,6 +129,12 @@ export class SearchStore {
 				}
 			},
 		);
+
+		reaction(
+			() => this.searchForm.parentEvent,
+			parentEvent =>
+				parentEvent && !isEventId(parentEvent) && this.loadEventAutocompleteList(parentEvent),
```
rzvdv, line 136: clear autocompleteList if there is no parent event.
Event was copy/paste, but since we now accept event name i think we should debounce it

src/components/filter/row/EventResolverRow.tsx
```diff
@@ -120,17 +138,27 @@ export default function EventResolverRow({ config }: { config: FilterRowEventRes
 					)}
 				</div>
 			)}
-			<div className={searchStatusIconClassname} />
-			<input
-				type='text'
+			{(isLoading || isError) && <div className={searchStatusIconClassname} />}
```
rzvdv, line 141: After finding event by id we show event data: name, timestamp and status. But now if you find event by id and you focus input and then blur this data doesn't show up

src/stores/SearchStore.ts
```diff
@@ -688,6 +706,38 @@ export class SearchStore {
...
 	}
 
+	@action resetEventAutocompleteList = () => {
+		this.eventAutocompleteList = [];
+	};
+
+	private loadEventAutocompleteList = debounce((parentEventName: string) => {
+		if (this.eventAutocompleteSseChannel) {
+			this.eventAutocompleteSseChannel.close();
+			this.resetEventAutocompleteList();
```
rzvdv, line 716: i think that's another bug.
this line should be outside of if statement. otherwise previos results stay in the list and the new ids append to the end of it

rzvdv (approved):

PR #254, opened 2021-09-30
Th2-2367 Implement bulk message export

molotgor:


src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -282,6 +284,15 @@ const MessagesFilterPanel = () => {
 			<MessageReplayModal />
 			<MessageFilterWarning />
 			<MessagesFilterSessionFilter config={sessionFilterConfig} submitChanges={submitChanges} />
+			{true && (
```
rzvdv, line 287: This was supposed to be something else then `true` i guess

src/components/message/message-card/MessageCard.tsx
```diff
@@ -149,7 +151,9 @@ const MessageCard = observer(({ message, viewType, setViewType }: Props) => {
 			isSoftFiltered={isSoftFiltered}
 			toogleMessagePin={toogleMessagePin}
 			isDetailed={isDetailed}
+			isExported={!!isExported}
```
rzvdv, line 154: It does not make sense to use double NOT operator for boolean

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -282,6 +284,15 @@ const MessagesFilterPanel = () => {
...
+				<MessageExport
+					isExport={messagesStore.isExport}
+					enableExport={() => messagesStore.enableExport()}
```
rzvdv, line 290: you can pass a function reference directly without wrapping it with another function

`enableExport={messagesStore.enableExport}`

src/stores/messages/MessagesStore.ts
```diff
@@ -128,6 +135,88 @@ export default class MessagesStore {
 		return [timestampTo - 15 * 1000, timestampTo + 15 * 1000];
 	}
 
+	public includes(message: EventMessage) {
```
rzvdv, line 138: this is very confusing when you use names of built in methods.
`messagesStore.includes(message)` looks likes messagesStore is an array

rzvdv (changes requested):

src/stores/messages/MessagesStore.ts
```diff
@@ -49,6 +50,12 @@ export default class MessagesStore {
 
 	public dataStore = new MessagesDataProviderStore(this, this.api);
 
+	@observable
```
rzvdv, line 53: let's make MessagesExportStore or something like that to keep MessagesStore as thin as possible

src/stores/messages/MessagesStore.ts
```diff
@@ -128,6 +135,88 @@ export default class MessagesStore {
 		return [timestampTo - 15 * 1000, timestampTo + 15 * 1000];
 	}
 
+	public includes(message: EventMessage) {
+		return (
+			this.exportMessages.filter(exportMessage => exportMessage.messageId === message.messageId)
```
rzvdv, line 140: i think we can even check it by reference and not by id, and replace filter with includes

src/stores/messages/MessagesStore.ts
```diff
@@ -128,6 +135,88 @@ export default class MessagesStore {
...
+		}
+
+		return content;
+	}
+
+	@action
+	public endExport(messageViewType: MessageViewType) {
+		this.isExport = false;
+		if (this.exportMessages.length === 0) return;
+		const exportUrl = URL.createObjectURL(
```
rzvdv, line 200: there is downloadTxtFile function which currently isn't used anywhere, you can modify it and move all file creating-downloading logic there

src/stores/workspace/WorkspaceStore.ts
```diff
@@ -158,6 +158,7 @@ export default class WorkspaceStore {
 	@action
 	public onSavedItemSelect = (savedItem: EventTreeNode | EventAction | EventMessage) => {
 		if (isEventMessage(savedItem)) {
+			this.messagesStore.disableExport();
```
rzvdv, line 161: Maybe we should reset export on messages filter change?

src/stores/messages/MessagesStore.ts
```diff
@@ -128,6 +135,88 @@ export default class MessagesStore {
...
+					this.exportMessages
+						.map(exportMessage => this.convertMessage(exportMessage, messageViewType))
+						.join('\n'),
+				],
+				'exported_messages.txt',
+				{ type: 'text/plain;charset=UTF-8;base64' },
+			),
+		);
+		const link = document.createElement('a');
+		link.download = 'exported_messages.txt';
```
rzvdv, line 212: let's add date-time to a filename

src/components/filter/MessagesFilterPanel.tsx
molotgor: If I don't wrap it then there is error because 'this' in 'enableExport' and 'disableExport' is 'undefined'.

rzvdv: That's why i prefer using arrow functions inside stores as well. if you convert disableExport, enableExport, endExport to be arrow functions you can avoid wrapping them

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -282,6 +284,17 @@ const MessagesFilterPanel = () => {
 			<MessageReplayModal />
 			<MessageFilterWarning />
 			<MessagesFilterSessionFilter config={sessionFilterConfig} submitChanges={submitChanges} />
+			{
```
rzvdv, line 287: Delete brackets

src/stores/messages/MessagesStore.ts
```diff
@@ -264,6 +270,7 @@ export default class MessagesStore {
 		this.selectedMessageId = null;
 		this.highlightedMessageId = null;
 		this.hintMessages = [];
+		this.exportStore.disableExport();
```
rzvdv, line 273: we can delete this line since reaction to `filterStore.filter` will be triggered

src/stores/messages/MessagesStore.ts
```diff
@@ -279,6 +286,7 @@ export default class MessagesStore {
 	@action
 	public clearFilters = () => {
 		this.hintMessages = [];
+		this.exportStore.disableExport();
```
rzvdv, line 289: and also delete that

rzvdv (commented):
Can we add some styles when export is enabled. Maybe add dashed border to message and cursor pointer so user can understand that he can select messages?

rzvdv (approved):

rzvdv (dismissed):

PR #255, opened 2021-10-01
TH2-2461. Make the messages list to react when display mode changes.

VitalyMashanov:
* fixed but with a problem. need to discuss.


src/stores/messages/SavedMessageViewType.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************** */
+
+import { action, observable } from 'mobx';
+import { MessageViewType } from '../../models/EventMessage';
```
rzvdv, line 18: I think thats good idea to delete SaveSaver and use class for managing message viewType, but i want suggest some changes:

i think it should look something like this:
```
class MessageViewStore {
   constructor(public messageId: string, public session: string) {
         reaction(() => this.displayRule) // here we can react to change of rule and change view message view type
   }
   
   @computed displayRule

   viewType // will default to displayRule.viewType

   setViewType

}

```

src/stores/messages/SavedMessagesViewTypesStore.ts
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************** */
+
+import { action, observable } from 'mobx';
+import { SavedMessageViewType } from './SavedMessageViewType';
+
+class SavedMessagesViewTypesStore {
```
rzvdv, line 20: i think this store should have instance per workspace, not global, also after messages full reload this has to be cleared

rzvdv (changes requested):

src/stores/messages/SavedMessagesViewTypesStore.ts
```diff
@@ -0,0 +1,44 @@
...
+	@action
+	public saveViewType = (id: string, vt: SavedMessageViewType) => {
+		if (!this.savedViewTypes.has(id)) {
+			this.savedViewTypes.set(id, vt);
+		}
+	};
+
+	@action
+	public getSavedViewType = (id: string) => {
+		return this.savedViewTypes.get(id);
```
rzvdv, line 33: and if viewStore wasnt initiated and its undefined we can create an instance of ViewStore, and return it

PR #256, opened 2021-10-04
[TH2-2456][ReportViewerUI] Main input doesn't perform search by ID

molotgor:


rzvdv (changes requested):
update version in package.json not in package-lock.json
Also disable spellcheck for GraphSearchInput please

rzvdv (approved):

PR #257, opened 2021-10-05
SSELoaders refactoring

VekesserVladimir:
Refactoring SSE loaders. All sse api calls are replaced with loaders where possible


src/stores/events/EventsDataStore.ts
```diff
@@ -261,7 +261,7 @@ export default class EventsDataStore {
 			}
 		} catch (error) {
 			console.error(error);
-			if (error.name !== 'AbortError') {
+			if (error instanceof DOMException && error.code === error.ABORT_ERR) {
```
rzvdv, line 264: it was checking if it isn't an AbortError, but your code does the opposite.
i think we can also extract this check to a function isAbortError since we use it often

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -405,7 +404,7 @@ export default class MessagesDataProviderStore {
 			});
 		} catch (error) {
 			runInAction(() => {
-				if (error.name !== 'AbortError') {
+				if (error instanceof DOMException && error.code === error.ABORT_ERR) {
```
rzvdv, line 407: same here, should check if it's not an abort error

src/stores/SSEChannel/SSEChannel.ts
```diff
@@ -1,199 +1,161 @@
...
+
+	protected initialResponseTimeout: number | null = null;
+
+	private updateScheduler: number | null = null;
+
+	protected fetchedChunkSubscription: WhenPromise | null = null;
+
+	private updateSchedulerIntervalMs = 1000;
+
+	protected fetchedEventsCount = 0;
```
rzvdv, line 60: looks like we don't need this property

src/helpers/fetch.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************** */
+
+export const isAbortError = (error: unknown): boolean => {
+	return error instanceof DOMException && error.code !== error.ABORT_ERR;
```
rzvdv, line 18: if this is isAbortError function shouldn't it return error.code == error.ABORT_ERR ?

PR #278, opened 2021-11-09
small-inputs-fixes

VitalyMashanov:
* graph search input value trimming
 * autocomplete list show fix


src/components/graph/search/GraphSearchInput.tsx
```diff
@@ -225,7 +225,7 @@ function GraphSearchInput(props: Props) {
 	}
 
 	function handleInputValueChange(event: React.ChangeEvent<HTMLInputElement>) {
-		const value = event.target.value;
+		const value = event.target.value.trim();
```
rzvdv, line 228: I think there is still feature that allows to enter date, and while entering date e.g `02.11.2021 00:00:00.000` space is required

src/components/util/AutocompleteInput/AutocompleteList.tsx
```diff
@@ -157,9 +157,7 @@ export const AutocompleteList = React.forwardRef<HTMLDivElement, AutocompleteLis
 		}, [handleKeyDown]);
 
 		React.useEffect(() => {
-			const showAutocomplete = Boolean(
-				(alwaysShow || (anchor && value && value !== focusedOption)) && list.length,
-			);
+			const showAutocomplete = Boolean((alwaysShow || (anchor && value)) && list.length);
```
rzvdv, line 160: why was this change required?
There is issue now when there is only one item in autocomplete list, you try to click it, and autocomplete list is still showing. But i would expect it to close

src/components/graph/search/GraphSearchInput.tsx
VitalyMashanov: i think trim method is not for getting rid of all spaces but cuts off string's end and start spaces only...
correct me if it is not or clarify what case do you mean

src/components/util/AutocompleteInput/AutocompleteList.tsx
VitalyMashanov: that s why
image

src/components/graph/search/GraphSearchInput.tsx
rzvdv: try to enter `02.11.2021 00:00:00.000` . you wont be able to insert space after date

src/components/util/AutocompleteInput/AutocompleteList.tsx
```diff
@@ -158,8 +158,11 @@ export const AutocompleteList = React.forwardRef<HTMLDivElement, AutocompleteLis
 
 		React.useEffect(() => {
 			const showAutocomplete = Boolean(
-				(alwaysShow || (anchor && value && value !== focusedOption)) && list.length,
+				(alwaysShow ||
+					(anchor && value && focusedOption?.toLocaleLowerCase().includes(value.toLowerCase()))) &&
```
rzvdv, line 162: not sure why `toLocaleLowerCase` is used.

Autocomplete list accepts anchor, so what if we setAnchor to null upon selecting value, i think this might solve the problem

rzvdv (approved):

PR #279, opened 2021-11-12
message-card-tools-restyling

VitalyMashanov:


src/components/message/message-card/MessageCardTools.tsx
```diff
@@ -64,10 +64,10 @@ const MessageCardTools = ({
 
 	const viewTypes = message.body
 		? [
-				MessageViewType.JSON,
 				MessageViewType.FORMATTED,
```
rzvdv, line 67: why has order changed? i think json and formatted view types should be next to each other

src/components/message/message-card/MessageCardTools.tsx
```diff
@@ -172,22 +160,16 @@ const MessageCardTools = ({
 					(messageViewType === MessageViewType.JSON ||
 						messageViewType === MessageViewType.FORMATTED) && (
 						<div className='message-card-tools__controls-group'>
-							{['body', 'fields'].map(copyOption => (
+							{['fields'].map(copyOption => (
```
rzvdv, line 163: we don't need body copying anymore? if that's so array with 1 item looks weird we should refactor that

rzvdv (changes requested):
Pick any message that fits in one line, switch to ASCII mode, option window now has scrollbar

src/components/message/message-card/MessageCardTools.tsx
VitalyMashanov: according to https://www.figma.com/file/vB9EyLv1GBE3vJqy2Hxk08/TH2-New-Report-Style-Updates?node-id=2738%3A85943

rzvdv: i think previous order makes more sense. let's leave it as it was

rzvdv (changes requested):
Either we a make temporary solution for copying message without deleting part of functionality or we wait until sketch is updated, otherwise i don't think it should be merged

PR #281, opened 2021-11-18
[TH2-2653]. Fix messages list scrolls to attached message after next/prev chunk is loaded

rzvdv:
…prev chunk loaded


src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -75,7 +68,10 @@ const MessagesVirtualizedList = (props: Props) => {
 	React.useEffect(() => {
 		if (scrolledIndex !== null) {
 			raf(() => {
-				virtuoso.current?.scrollToIndex({ index: scrolledIndex.valueOf(), align: 'center' });
+				virtuoso.current?.scrollToIndex({
+					index: scrolledIndex,
+					behavior: 'smooth',
```
rzvdv, line 73: When i tested it worked not stable, i don't think we need that

src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -30,14 +30,7 @@ interface Props {
...
 	itemRenderer: (index: number, message: EventMessage) => React.ReactElement;
-	/*
-		Number objects is used here because in some cases (eg one message / action was 
-		selected several times by different entities)
-		We can't understand that we need to scroll to the selected entity again when
-		we are comparing primitive numbers.
-		Objects and reference comparison is the only way to handle numbers changing in this case.
-	*/
-	scrolledIndex: Number | null;
+	scrolledIndex: number | null;
```
rzvdv, line 33: It should be a Number object. There are some edge cases when previous scrolledIndex equals to new scrolledIndex, so if it was a number primivite type we couldn't detect it

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -137,6 +137,12 @@ export default class MessagesDataProviderStore {
 				}
 			}
 
+			const { selectedMessageId, scrollToMessage } = this.messagesStore;
+			this.searchChannelNext.onStop = () =>
```
rzvdv: this code should be deleted

VitalyMashanov: when you're picking some event, message list is not scrolling to attached message without this part

src/components/message/message-card-list/MessagesVirtualizedList.tsx
VitalyMashanov: when you choose another event with the same attached message
messages list is reloading
is there any other edge cases?

src/stores/messages/MessagesDataProviderStore.ts
VitalyMashanov: for example

rzvdv: then we should develop new appropriate solution to scroll to selected message

rzvdv:
fixed and merged in #275

rzvdv:
fixed and merged in #275

PR #282, opened 2021-11-26
Fix MobX warnings.

artem-davidovich:


src/stores/SearchStore.ts
```diff
@@ -113,7 +113,7 @@ export class SearchStore {
 		this.init();
 
 		autorun(() => {
-			this.currentSearch = this.searchHistory[this.currentIndex] || null;
+			this.currentSearch = this.currentIndex > 0 ? this.searchHistory[this.currentIndex] : null;
```
rzvdv, line 116: What type of warning does it fix?
Why current index has to be greater than 0 ? we ignore 0 index element?

that warning?
Attempt to read an array index (0) that is out of bounds (0). 

I think it's about checking array length first before accessing it.

rzvdv (approved):

PR #283, opened 2021-11-26
updated node version in docker

molotgor:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10.24 AS build
+FROM node:15.14 AS build
```
rzvdv, line 1: I think even versions of node are LTS, so i'd rather go for 14 or 12

rzvdv (approved):

PR #286, opened 2021-12-01
TH2-2754

VitalyMashanov:
[ReportViewerUI] Focus always returns to the selected message attached to event after pressing the [keep loading] button

 * fixed


rzvdv:
duplicate of  #281

PR #297, opened 2021-12-11
TH2-2837. Fixed filter layout

MariaMakii:


rzvdv (changes requested):
But this wont allow values to take multiple rows, maybe if value is too long we can make filter panel wider or there is other solution to this?

PR #300, opened 2021-12-14
message-header-restyling

VitalyMashanov:


src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -68,25 +69,40 @@ export function MessageCardBase({
...
+		isAttached ? 'attached' : null,
+		isBookmarked ? 'pinned' : null,
+	);
+
 	const renderInlineMessageInfo = () => {
 		if (viewType === MessageViewType.ASCII || viewType === MessageViewType.JSON) {
 			const formattedTimestamp = formatTime(timestampToNumber(timestamp));
+			const splittedMessageId = messageId.split(':');
+			const messageIdWithHighlightedSession = splittedMessageId.map(
+				splittedMessageId.length > 3
```
rzvdv, line 83: instead of using this ternary with 2 functions, your function can accept array which .map provides 3rd argmunent

src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -68,25 +69,40 @@ export function MessageCardBase({
...
+					? extendedGetMessageIdWithHighlightedSession
+					: getMessageIdWithHighlightedSession,
+			);
 			return (
 				<>
+					{isExport && (
+						<Checkbox
```
rzvdv, line 90: This checkbox shifts whole message list, i think we need to avoid it

src/styles/common/_colors.scss
```diff
@@ -209,3 +209,8 @@ $bookmarkPanelButtonBackgroundColor: #cca3f5;
 $bookmarkLinkHoverColor: #040505;
 $eventsListLoadButtonColor: #e5eaf0;
 $eventsListLoadButtonTextColor: #bc271a;
+
+$attachedMessageCardColor: #FFA64C;
```
rzvdv, line 213: there is alread attachedMessageColor and other color variables from this one. Can we check if they are even used anywhere? if not let's delete then, otherwise its confusing that we have 2 attached message color variables

src/components/message/message-card/MessageCardBase.tsx
VitalyMashanov: it is added according to https://www.figma.com/file/vB9EyLv1GBE3vJqy2Hxk08/TH2-New-Report-Style-Updates?node-id=2738%3A80177

PR #304, opened 2021-12-22
TH2-2370: Tree table card cannot show more than 62 groups

VekesserVladimir:


src/styles/tables.scss
```diff
@@ -439,7 +365,10 @@
 
 .params-table {
 	@include default-table;
-	@include scrollbar;
+
+	&-wrapper {
+		overflow-x: scroll;
```
rzvdv, line 370: isnt auto a better option?

src/styles/common/_mixins.scss
```diff
@@ -203,7 +202,6 @@
 		background-color: $tableBodyBackground;
 		border-right: 0;
 		border-left: 0;
```
rzvdv, line 204: i think this breaks aligning in first column of verification table

rzvdv (changes requested):

src/styles/common/_mixins.scss
VekesserVladimir: Fixed

src/styles/tables.scss
VekesserVladimir: Fixed

rzvdv (approved):

PR #307, opened 2021-12-28
message-card-tools-restyling

VitalyMashanov:


src/components/message/message-card/MessageCardTools.tsx
```diff
@@ -175,19 +161,13 @@ const MessageCardTools = ({
 							{['body', 'fields'].map(copyOption => (
```
rzvdv, line 161: There should be only 1 copy button for 'body', remove fields

rzvdv (changes requested):
When message is a one-liner MessageCardTools height is bigger than MessageCard and it creates overflow

PR #309, opened 2021-12-28
message-restyling

VitalyMashanov:


src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -48,168 +49,201 @@ export interface MessageCardBaseProps {
...
+		);
+
+		const renderInlineMessageInfo = () => {
+			if (viewType === MessageViewType.ASCII || viewType === MessageViewType.JSON) {
+				const formattedTimestamp = formatTime(timestampToNumber(timestamp));
+				return (
+					<>
+						{isExport ? (
+							<Checkbox
+								id={messageId}
```
rzvdv, line 110: There should be stopPropagation() otherwise `addMessageToExport` will be called twice, in Checkbox component and in root div of MessageCardBase

rzvdv (changes requested):

src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -48,166 +49,200 @@ export interface MessageCardBaseProps {
...
+export default MessageCardBase;
 
 function calculateHueValue(session: string): number {
 	const hashCode = getHashCode(session);
 
 	return (hashCode % HUE_SEGMENTS_COUNT) * (360 / HUE_SEGMENTS_COUNT);
 }
+
+function getMessageIdWithHighlightedSession(part: string, i: number, arr: string[]) {
+	if (arr.length === 3) {
```
rzvdv, line 237: you still have to provide keys, right now message markup breaks after message is updated, try to switch export on and pick binary view type and then click the message.

this function returns jsx, maybe we should make it Session component then?

src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -48,166 +49,200 @@ export interface MessageCardBaseProps {
...
+			if (viewType === MessageViewType.FORMATTED || viewType === MessageViewType.BINARY) {
+				const formattedTimestamp = formatTime(timestampToNumber(timestamp));
+				return (
+					<div className='mc-header__info'>
+						{isExport ? (
+							<Checkbox
+								id={messageId}
+								checked={!!isExported}
+								onChange={e => {
+									e.stopPropagation();
```
rzvdv, line 175: this wont work, we have to stop click event propagation

src/components/message/message-card/MessageCardBase.tsx
VekesserVladimir: Markup breaking still actual

VekesserVladimir: Also we have the problem with the selection highlighting in a binary view mode

rzvdv (approved):

PR #310, opened 2021-12-28
TH2-2461. Make the messages list to react when display mode changes.

VitalyMashanov:
* fixed


rzvdv (changes requested):
There is legacy code in MessagesStore we can remove now:
beautifiedMessages
detailedRawMessagesIds
and methods that modify them

Also `resetSavedViewTypes` is not being called anywhere, i think we can add reaction and clear it on messages filter changes

PR #311, opened 2021-12-29
Fix 'AutocompleteInput' outside click.

artem-davidovich:


rzvdv (changes requested):
This doesn't work. Now when you select value from autocomplete list in MultipleStringFilterRow it adds 2 values.

PR #314, opened 2022-01-10
TH2-2753. Fix "Uncaught Error: Message with xxxyyyzzz id doesn't exists"

osinkinroman:


src/stores/messages/MessagesStore.ts
```diff
@@ -153,7 +153,8 @@ export default class MessagesStore {
 	public scrollToMessage = (messageId: string) => {
 		const messageIndex = this.dataStore.messages.findIndex(m => m.messageId === messageId);
 
-		if (messageIndex === -1) throw new Error(`Message with ${messageId} id doesn't exists`);
+		if (messageIndex === -1 && this.dataStore.messages.length > 1)
```
rzvdv, line 156: we can actually delete `scrollToMessage` and scrolledIndex

messageList scroll to message when `messagesStore.selectedMessageId` is changed

in MessagesUpdateStore replace `scrollToMessage` with assinging `messagesStore.selectedMessageId` new value

src/stores/messages/MessagesUpdateStore.ts
```diff
@@ -50,10 +50,10 @@ export default class MessagesUpdateStore {
...
 			},
 			{
-				onResponse: incommingMessages => {
-					if (incommingMessages.length) {
-						onNextChannelResponse(incommingMessages);
-						this.scrollToMessage(incommingMessages[0].messageId);
+				onResponse: incomingMessages => {
+					if (incomingMessages.length) {
+						onNextChannelResponse(incomingMessages);
+						this.selectedMessageId = incomingMessages[0].messageId;
```
rzvdv, line 56: change that to messageStore.selectedMessageId and remove `private selectedMessageId: String | null,`

rzvdv (approved):

PR #315, opened 2022-01-10
Messages loading bug and minor improvements

rzvdv:


src/components/search-panel/SearchPanelResults.tsx
```diff
@@ -51,16 +61,47 @@ const SearchPanelResults = (props: SearchPanelResultsProps) => {
...
+					filters={filters}
+					onResultClick={onResultItemClick}
+					onGroupClick={onResultGroupClick}
+				/>
+			</React.Fragment>
+		);
+	};
+
+	const loadMoreButton = () => {
+		return !loadMoreButton ? null : (
```
rzvdv, line 98: is that supposued to be `showLoadMoreButton` ?

src/components/search-panel/SearchPanelResults.tsx
```diff
@@ -51,16 +61,47 @@ const SearchPanelResults = (props: SearchPanelResultsProps) => {
...
 	}
 
+	const isSeparator = (object: FlattenedResult): object is Separator => {
+		return !Number.isNaN(+object[0]);
+	};
+
+	const renderResult = (index: number, results: FlattenedResult) => {
+		if (isSeparator(results)) {
+			return (
+				<React.Fragment>
```
rzvdv, line 80: i dont think fragments should be used here

PR #319, opened 2022-01-10
added loading screen to fallback

molotgor:


rzvdv (approved):

src/styles/root.scss
```diff
@@ -136,3 +136,15 @@ input {
 		background-color: $graphOverlayFailedEventsColor;
 	}
 }
+
+.loading {
```
rzvdv, line 140: I think there some .loading classes already

rzvdv (changes requested):
I think there are some .loading classes already

rzvdv (approved):

PR #321, opened 2022-01-11
updated 6.1-based filter styles

Ryzhkov-Alexey:


src/components/filter/row/TogglerRow.tsx
```diff
@@ -37,9 +37,9 @@ const TogglerRow = ({ config }: { config: FilterRowTogglerConfig }) => {
 						toggleValue();
 					}
 				}}>
-				<p className='toggler__label'>{firstLabel}</p>
+				<p className={`toggler__label ${value ? 'selected' : ''}`}>{firstLabel}</p>
```
rzvdv, line 40: maybe you can apply class to `togglerClassName` and then use selectors to target first or last p tag?

src/styles/filter.scss
```diff
@@ -506,7 +510,7 @@
 
 	&__label {
 		width: 120px;
-		font-weight: bold;
+		font-weight: 600;
```
rzvdv, line 513: when font weight is being changed it actually makes the label a bit wider, and it moves the rest of the filter row, can you fix it?

rzvdv (changes requested):

PR #322, opened 2022-01-12
[6.1] Workspaces updates

icefleen:


src/components/bookmarks/BookmarksPanel.tsx
```diff
@@ -200,6 +199,9 @@ export default observer(BookmarksPanel);
 
 interface BookmarkItemProps {
 	bookmark: BookmarkedItem;
+	messages?: EventMessage[];
```
rzvdv, line 202: i cant see these props being used anywhere, and passing list of messages to a bookmark doesnt seem like a good idea

src/components/App.tsx
```diff
@@ -18,21 +18,23 @@ import { hot } from 'react-hot-loader/root';
...
-import { useTabsStore } from '../hooks';
+import { useSessionsStore } from '../hooks';
 import '../styles/root.scss';
 
 const AppRootBase = () => {
-	const searchStore = useSearchStore();
-	const tabsStore = useTabsStore();
+	const sessionsStore = useSessionsStore();
+
+	const sessionNames = useMemo(() => sessionsStore.sessions.map(({ session }) => session), [
```
rzvdv, line 35: even though that's ok to use useMemo here, but wouldn't `computed` make more sense, since we are deriving it from observable array ?

src/stores/workspace/WorkspaceViewStore.ts
```diff
@@ -59,13 +62,165 @@ export default class WorkspaceViewStore {
...
-	public resetToDefaulLayout = () => {
-		this.panelsLayout = [50, 50];
+	public setCollapsedPanels = (collapsedPanels: number[]) => {
+		this.collapsedPanels = collapsedPanels;
 	};
 
 	@action
 	public collapsePanel = (index: number) => {
-		this.panelsLayout = index === 0 ? [0, 100] : [100, 0];
+		if (this.collapsedPanels.length === 0) {
```
rzvdv, line 71: do we really need `collapsedPanels` state? cant we just tell from panels width if its collapsed or not?
also algorithm looks rather complicated and long, i'm sure there must be a better way to do it

rzvdv:
```
		const layouts = [[100], [50, 50], [35, 35, 30], [35, 25, 20, 20]];
		const notCollapsedPanels = this.panelsLayout
			.map((w, i) => (w > 10 ? i : null))
			.filter(w => w !== null);
		const isTargetPanelCollapsed = this.panelsLayout[index] < 10;
		const panelCount = notCollapsedPanels.includes(index)
			? notCollapsedPanels.length - 1
			: notCollapsedPanels.length + 1;

		if (panelCount <= 0) return;
		const widths = layouts[panelCount - 1];

		const updatedLayout = [];
		for (let i = 0; i < this.panelsLayout.length; i++) {
			let width: number | undefined = 0;
			if (i === index) {
				width = isTargetPanelCollapsed ? widths.shift() : 0;
			} else if (this.panelsLayout[i] > 10) {
				width = widths.shift();
			}
			updatedLayout.push(width || 0);
		}

		this.panelsLayout = updatedLayout as any;
```

src/components/App.tsx
icefleen: fixed: https://github.com/th2-net/th2-rpt-viewer/pull/322/commits/fe67f8dcf54fe187157b865a90648d7ea6df6574

src/components/bookmarks/BookmarksPanel.tsx
icefleen: that was a logic about highlighting elements if they're exist in events or messages panel. Removed from bookmark components and implemented in actual components: https://github.com/th2-net/th2-rpt-viewer/pull/322/commits/61548c3c2c7f81cf33a68c8a83f8b860fbffee13

src/components/App.tsx
```diff
@@ -24,15 +24,12 @@ import ToastContainer from './notifications/ToastContainer';
...
 import MessagesViewConfigurator from './messages-view-configurator/MessagesViewConfigurator';
-import { useSearchStore } from '../hooks/useSearchStore';
-import { useTabsStore } from '../hooks';
+import { useSessionsStore } from '../hooks';
 import '../styles/root.scss';
 
 const AppRootBase = () => {
-	const searchStore = useSearchStore();
-	const tabsStore = useTabsStore();
+	const { sessionNames } = useSessionsStore();
```
rzvdv, line 32: wait i think we confused something. sessionStore keeps sessions submitted by the user, and searchStore fetches all available sessions, and MessagesViewConfigurator waits for all sessions

rzvdv (approved):

PR #325, opened 2022-01-17
TH2-2959 - General root group for all unknown events

VekesserVladimir:


src/components/event/EventList.tsx
```diff
@@ -126,6 +126,10 @@ function EventTreeListBase(props: Props) {
 		}
 	}, [scrolledIndex]);
 
+	React.useEffect(() => {
+		initialItemCount.current = nodes.length;
```
rzvdv, line 130: I'm not sure why do we need this?

This property applies only on mount and it's default value declared in initialItemCount ref should be enough

src/helpers/event.ts
```diff
@@ -131,25 +131,19 @@ export const convertEventActionToEventTreeNode = (event: EventAction): EventTree
...
-		},
-		successful: false,
-	};
+export const unknownRoot: EventTreeNode = {
+	type: ActionType.EVENT_TREE_NODE,
+	isUnknown: true,
+	eventId: 'unknown-root',
+	eventName: 'Unknown Events',
+	eventType: 'eventTreeNode',
+	filtered: false,
```
rzvdv, line 140: it looks like `filtered` is not used anywhere, maybe we should delete it?

src/components/event/EventWindow.tsx
```diff
@@ -33,7 +33,9 @@ function EventWindow() {
 		if (node) {
 			eventsStore.scrollToEvent(node.eventId);
 		}
-		eventsStore.selectNode(node);
+		if (!node?.isUnknown) {
```
rzvdv, line 36: id rather move this check inside of `eventsStore.selectNode` instead of adding everywhere

src/stores/events/EventsDataStore.ts
```diff
@@ -92,10 +88,19 @@ export default class EventsDataStore {
 	@computed
 	public get targetNodePath() {
 		if (!this.targetNode) return null;
-		const parentIds = this.eventStore
-			.getParentNodes(this.targetNode.eventId, this.eventsCache)
-			.map(parentEventNode => parentEventNode.eventId);
-		return [...parentIds, this.targetNode.eventId];
+
```
rzvdv, line 91: is there a reason it had to be rewritten? looks like it reimplements `eventStore.getParentNodes`.

src/stores/events/EventsDataStore.ts
```diff
@@ -165,7 +170,10 @@ export default class EventsDataStore {
 			if (event.parentId === null) {
 				rootEvents.push(event);
 			} else {
-				eventsByParentId[event.parentId] = (eventsByParentId[event.parentId] || []).concat(event);
+				const cachedParent = this.eventsCache.get(event.parentId);
+				// We need this to change the parent to an unknown root if needed
+				const parentId = cachedParent?.eventId ?? event.parentId;
```
rzvdv, line 175: if `cachedParent.eventId` and `event.parentId` is the same string, why do we need to even get parent from cache and check it?

src/stores/events/EventsDataStore.ts
```diff
@@ -284,14 +292,34 @@ export default class EventsDataStore {
...
 					} else {
-						rootNode = getErrorEventTreeNode(rootNode?.parentId || parentId);
+						rootNode = unknownRoot;
+						if (currentParentId) {
+							this.parentChildrensMap.set(unknownRoot.eventId, [
+								...(this.parentChildrensMap.get(unknownRoot.eventId) || []),
+								...(this.parentChildrensMap.get(currentParentId) || []),
+							]);
+							this.parentChildrensMap.delete(currentParentId);
+							this.rootEventIds = [...this.rootEventIds, currentParentId];
```
rzvdv, line 302: why currentParentId should be in `rootEventIds`? isn't that an uknown event that supposed to be in `unknownRoot`

src/stores/events/EventsDataStore.ts
```diff
@@ -284,14 +292,34 @@ export default class EventsDataStore {
...
 					if (!this.eventsCache.has(eventNode.eventId)) {
-						this.eventsCache.set(eventNode.eventId, eventNode);
+						if (eventNode.eventId !== unknownRoot.eventId) {
+							this.eventsCache.set(eventNode.eventId, eventNode);
+						} else {
+							const eventsMap: Map<string, EventTreeNode> = observable.map(
+								new Map<string, EventTreeNode>([
+									...this.eventsCache,
+									// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+									[currentParentId!, unknownRoot],
```
rzvdv, line 317: There should be only one `unknownRoot` which contains another unknown events, we might want to distinguish them. So maybe besides `unknownRoot` we should have `unknownEventNode`

src/stores/events/EventsDataStore.ts
```diff
@@ -317,21 +343,24 @@ export default class EventsDataStore {
...
 
 		parentNodes.forEach(parentNodePath => {
 			for (let i = 0; i < parentNodePath.length; i++) {
 				const event = parentNodePath[i];
 
-				const { isUnknown, parentId, eventId } = event;
+				const { parentId: realParentId, eventId } = event;
+				const cachedParent = realParentId ? this.eventsCache.get(realParentId) : null;
+				// We need this to change the parent to an unknown root if needed
+				const parentId = cachedParent?.eventId ?? event.parentId;
```
rzvdv, line 361: same problem, i don't understand why do we need to get parent event from cache, if in the end it will be `event.parentId`

src/stores/events/EventsStore.ts
```diff
@@ -161,17 +167,22 @@ export default class EventsStore {
...
 				rootIds.push(rootNode.eventId);
 			}
 		}
 
-		const rootNodes = sortEventsByTimestamp(
-			rootIds.map(eventId => this.eventDataStore.eventsCache.get(eventId)).filter(isEventNode),
-			'desc',
-		);
+		const rootNodes = [
+			...new Set(
```
rzvdv, line 179: I thinks it's better not to add duplicate id's then filter it afterwards. Why do we have duplicates there? unknownRoot is added multiple times?

src/helpers/event.ts
VekesserVladimir: `filtered` is a required property in 'EventTreeNode'

rzvdv: is it ? i can't find any reference to `.filtered` as far as i remember this property is from old api

src/stores/events/EventsDataStore.ts
VekesserVladimir: Here we are accumulating event by id they was searched. Cache contains unknown group if event wasn't found, but we need real ids

src/stores/events/EventsDataStore.ts
VekesserVladimir: `CachedParent.eventId` may be `'unknown-root'` because cache contains unknown group if event wasn't found

src/stores/events/EventsStore.ts
VekesserVladimir: Because we have only real ids in `eventDataStore.rootEventIds` we can get several unknown group nodes.

src/stores/events/EventsDataStore.ts
VekesserVladimir: We contains only real ids eventDataStore.rootEventIds for correct target node resolving

src/helpers/event.ts
VekesserVladimir: Ok, I deleted it

src/stores/events/EventsDataStore.ts
```diff
@@ -284,14 +292,34 @@ export default class EventsDataStore {
 					if (cachedRootNode) {
 						rootNode = cachedRootNode;
 					} else {
-						rootNode = getErrorEventTreeNode(rootNode?.parentId || parentId);
+						rootNode = unknownRoot;
+						if (currentParentId) {
```
rzvdv, line 296: I wonder if our understanding of this task is different. Suppose we have 3 events, 2 of them are unknown
I expect tree to look like this:
-- event1
-- unknownRoot
   ---- event2
        ---- event2 children
   ---- event3
        ---- event3 children

but you set all the children of unknown event to unknownRoot

that's why i said we need another not root unknown event

src/components/event/EventList.tsx
```diff
@@ -133,7 +133,7 @@ function EventTreeListBase(props: Props) {
 			<StateSaverProvider>
 				<Virtuoso
 					firstItemIndex={firstItemIndex}
-					initialTopMostItemIndex={initialItemCount.current - 1}
+					initialTopMostItemIndex={initialItemCount.current ? initialItemCount.current - 1 : 0}
```
rzvdv, line 136: that's not necessary since event list will not even render if list is empty

src/stores/events/EventsDataStore.ts
rzvdv: I think that's confusing, i think there should be only real root event id's. There might be some diffucilties with resolving and expanding target node if it's inside the unknown event, but i don't think that's a good solution to it

rzvdv (approved):

PR #328, opened 2022-01-20
TH2-3021. [ReportViewerUI] 'Copy full/Copy simplified' buttons disappear from panel

VitalyMashanov:
* fixed


rzvdv (changes requested):
Update version

rzvdv (approved):

PR #330, opened 2022-01-20
TH2-3008: Stop message loading button

VekesserVladimir:
Added stop message loading button
Removed loading spinner from filters button in filters panel

### DIFF
--- a/src/components/filter/FilterPanel.tsx
+++ b/src/components/filter/FilterPanel.tsx
@@ -27,7 +27,6 @@ const PANEL_WIDTH = 840;
 
 interface Props {
 	isFilterApplied: boolean;
-	isLoading?: boolean;
 	isDisabled?: boolean;
 	showFilter: boolean;
 	config: FilterRowConfig[];
@@ -47,7 +46,6 @@ const FilterPanel = (props: Props) => {
 		onSubmit,
 		onClearAll,
 		renderFooter,
-		isLoading,
 	} = props;
 
 	const filterBaseRef = React.useRef<HTMLDivElement>(null);
@@ -124,9 +122,6 @@ const FilterPanel = (props: Props) => {
 			<div className={filterButtonClass} ref={filterButtonRef} onClick={onClick}>
 				<div className={filterIconClass} />
 				<div className={filterTitleClass}>{showFilter ? 'Hide Filter' : 'Show Filter'}</div>
-				{typeof isLoading === 'boolean'
-					? isLoading && <div style={{ marginLeft: 5 }} className='filter__loading' />
-					: null}
 			</div>
 			<ModalPortal isOpen={showFilter}>
 				<div ref={filterBaseRef} className='filter'>
--- a/src/components/filter/FilterPanel.tsx
+++ b/src/components/filter/FilterPanel.tsx
@@ -27,6 +27,7 @@ const PANEL_WIDTH = 840;
 
 interface Props {
 	isFilterApplied: boolean;
+	isLoading: boolean;
 	isDisabled?: boolean;
 	showFilter: boolean;
 	config: FilterRowConfig[];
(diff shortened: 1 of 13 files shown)


src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -100,7 +100,8 @@ export default class MessagesDataProviderStore {
 		return (
 			this.isLoadingNextMessages ||
 			this.isLoadingPreviousMessages ||
-			Boolean(this.anchorChannel?.isLoading)
+			Boolean(this.anchorChannel?.isLoading) ||
+			Boolean(this.messageAC)
```
rzvdv, line 104: messageAC isn't observable
so this computed property will not react when it's changed
maybe we should add `isLoadingAnchorMessage` state?

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -109,6 +110,7 @@ export default class MessagesDataProviderStore {
 	@action
 	public loadMessages = async () => {
 		this.stopMessagesLoading();
+		this.resetMessagesDataState();
```
rzvdv, line 113: since we reseting state here now, can we delete this function call in `onFilterChange`

src/components/filter/FilterPanel.tsx
```diff
@@ -124,9 +122,6 @@ const FilterPanel = (props: Props) => {
 			<div className={filterButtonClass} ref={filterButtonRef} onClick={onClick}>
 				<div className={filterIconClass} />
 				<div className={filterTitleClass}>{showFilter ? 'Hide Filter' : 'Show Filter'}</div>
-				{typeof isLoading === 'boolean'
```
rzvdv, line 127: I see at least one scenario when having loading indicator is good.
If messages list contains messages, then you select some event, and while messages are loading(and they can all really long we had this problem) you should show some indicator that something is loading, right now if attached messages are loading stop loading button is active which is not right aswell

VekesserVladimir: While we loading attached messages we reset the messages list and have big spinner at the center of empty message list

rzvdv: No, we don't. Plus if filter is applied it will filter all attached messages, and they don't match current filter message list will not update

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -104,9 +106,16 @@ const MessagesFilterPanel = () => {
 		);
 	}, [filter, filterStore.filter, streams, isSoftFilterApplied]);
 
-	const isLoading = computed(
+	const stopLoading = React.useCallback(() => {
+		messagesDataStore.stopMessagesLoading();
+		workspaceStore.stopAttachedMessagesLoading();
+	}, []);
+
+	const isAnchorLoading = computed(() => workspaceStore.isLoadingAttachedMessages).get();
```
rzvdv, line 114: can we rename it to `isLoadingAttachedMessages`? or do we even need isAnchorLoading computed value ? can we just pass it to the component that's using it

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -104,9 +106,16 @@ const MessagesFilterPanel = () => {
...
+
+	const isMessageListLoading = computed(
 		() =>
-			(messagesDataStore.messages.length === 0 && messagesDataStore.isLoading) ||
+			messagesDataStore.isLoading ||
```
rzvdv, line 118: i think something is off with loading state, looks like it resets loading state before setting messages. And on initial loading there is some kind of bug, it shows that messages are loading, then it shows 'no messages' then spinner shows again

rzvdv (approved):

PR #333, opened 2022-01-24
[TH2-3054] Shorten workspace link

molotgor:
TH2-3054

### DIFF
--- a/src/components/filter/EventsFilterPanel.tsx
+++ b/src/components/filter/EventsFilterPanel.tsx
@@ -19,7 +19,6 @@ import { observer } from 'mobx-react-lite';
 import FilterPanel from './FilterPanel';
 import { FilterRowConfig, FilterRowTogglerConfig } from '../../models/filter/FilterInputs';
 import { useWorkspaceEventStore, useEventsFilterStore, useFiltersHistoryStore } from '../../hooks';
-import useEventsDataStore from '../../hooks/useEventsDataStore';
 import { EventSSEFilters } from '../../api/sse';
 import { Filter, EventFilterState } from '../search-panel/SearchPanelFilters';
 import { getObjectKeys, notEmpty } from '../../helpers/object';
@@ -28,6 +27,7 @@ import FiltersHistory from '../filters-history/FiltersHistory';
 import { getArrayOfUniques } from '../../helpers/array';
 import useSetState from '../../hooks/useSetState';
 import { prettifyCamelcase } from '../../helpers/stringUtils';
+import useEventsDataStore from '../../hooks/useEventsDataStore';
 
 type CurrentFilterValues = {
 	[key in EventSSEFilters]: string;
--- a/src/components/filter/EventsFilterPanel.tsx
+++ b/src/components/filter/EventsFilterPanel.tsx
@@ -45,6 +45,14 @@ function getDefaultCurrentFilterValues(filter: EventsFilter | null) {
 		: null;
 }
 
+const priority = [
+	'attachedEventIds-negative',
+	'type-include',
+	'name-include',
+	'body-include',
+	'status',
+];
+
 function EventsFilterPanel() {
 	const eventsStore = useWorkspaceEventStore();
 	const eventDataStore = useEventsDataStore();
@@ -56,14 +64,6 @@ function EventsFilterPanel() {
 	const [currentFilterValues, setCurrentFilterValues] = React.useState<CurrentFilterValues | null>(
 		getDefaultCurrentFilterValues(filterStore.filter),
 	);
-
-	const priority = [
-		'attachedEventIds-negative',
-		'type-include',
-		'name-include',
-		'body-include',
-		'status',
-	];
 
 	React.useEffect(() => {
 		setFilter(filterStore.filter);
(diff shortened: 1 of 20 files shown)


src/stores/RootStore.ts
```diff
@@ -73,9 +74,20 @@ export default class RootStore {
 		let messagesStoreState: MessagesStoreURLState = {};
 
 		if (activeWorkspace && isWorkspaceStore(activeWorkspace)) {
+			const clearEmpty = (filter: Partial<MessageFilterState> | Partial<EventsFilter>) => {
+				const tempFilter: Partial<MessageFilterState> | Partial<EventsFilter> = {};
```
rzvdv, line 78: i think you can just import toJS from mobx and replace these two lines with
`
	const tempFilter = toJS(filter)
`
it will convert observable to be plain js object

src/stores/RootStore.ts
```diff
@@ -73,9 +74,20 @@ export default class RootStore {
...
+				Object.assign(tempFilter, filter);
```
rzvdv, line 79: Also we can make some improvements by also deleting default values from url for example
`isSoftFilter` default value is false, so we can delete that
`flattenedListView` defaults to false

and also we can remove status filter from events if its value is `any`

rzvdv (changes requested):

rzvdv:
I noticed another problem, we don't have any sorting rules inside filter panel.
I think we need to have one. for example now if you fill filter status and name, they will be first in filter panel. and this looks weird, can you fix that aswell?

src/stores/RootStore.ts
```diff
@@ -73,35 +74,65 @@ export default class RootStore {
...
 				selectedEventId: eventsStore.selectedNode?.eventId,
-				flattenedListView: eventsStore.viewStore.flattenedListView,
+				flattenedListView:
+					eventsStore.viewStore.flattenedListView === false
+						? undefined
+						: eventsStore.viewStore.flattenedListView,
+			};
+			const clearMessageFilter = (filter: Partial<MessageFilterState> | Partial<EventsFilter>) => {
+				const tempFilter = toJS(filter);
+				Object.assign(tempFilter, filter);
```
rzvdv, line 105: Object.assign now can be deleted

src/components/filter/FilterPanel.tsx
```diff
@@ -130,17 +130,33 @@ const FilterPanel = (props: Props) => {
...
-								{rowConfig.map(_rowConfig => (
-									<FilterRow rowConfig={_rowConfig} key={_rowConfig.id} />
-								))}
-							</div>
-						) : (
-							<FilterRow rowConfig={rowConfig} key={rowConfig.id} />
-						),
-					)}
+					{config
+						.sort((a, b) => {
```
rzvdv, line 134: this is not exactly what i meant
i want events filter panel to have this order:
```
['attachedMessageId', 'type', 'name', 'body, 'text', 'status']
```
and for messages
```
['attachedEventIds', 'type', 'body', 'bodyBinary']
```

rzvdv (changes requested):

src/components/filter/FilterPanel.tsx
```diff
@@ -130,17 +130,44 @@ const FilterPanel = (props: Props) => {
...
+							const priority = [
```
rzvdv, line 135: can we move sorting to EventsFilterPanel and MessagesFilterPanel components? it'll be easier to sort it there and make more sense i think. also you can declare this order before component is defined, since it's constant, and we don't need to define it every time inside .sort compare function

src/components/filter/EventsFilterPanel.tsx
```diff
@@ -57,6 +57,14 @@ function EventsFilterPanel() {
 		getDefaultCurrentFilterValues(filterStore.filter),
 	);
 
+	const priority = [
```
rzvdv, line 60: can we declare it before EventsFilterPanel component ?

src/components/filter/MessagesFilterPanel.tsx
```diff
@@ -288,7 +295,21 @@ const MessagesFilterPanel = () => {
 				isFilterApplied={messagesStore.filterStore.isMessagesFilterApplied}
 				setShowFilter={setShowFilter}
 				showFilter={showFilter}
-				config={filterConfig}
+				config={filterConfig.sort((a, b) => {
```
rzvdv, line 298: can we sort it inside filterConfig useMemo?

src/components/filter/EventsFilterPanel.tsx
```diff
@@ -45,6 +45,14 @@ function getDefaultCurrentFilterValues(filter: EventsFilter | null) {
 		: null;
 }
 
+const priority = [
+	'attachedEventIds-negative',
```
rzvdv, line 49: attachedEventIds?  did you mean messages.
remove all this values and replace them with
keys of EventsFilter, same with Messages but there you will use keys of MessageFilterState

and sorting suppose to happen at line 125
`const filterNames = getObjectKeys(filter).sort(sortCb);`

rzvdv (dismissed):

rzvdv (approved):

PR #336, opened 2022-01-31
TH2-3063.  'Uncaught (in promise) WHEN_CANCELLED' error appears in console when click on Messages search several times

VitalyMashanov:
* workaround added.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -34,6 +34,19 @@ const AppRootBase = () => {
 	const searchStore = useSearchStore();
 	const tabsStore = useTabsStore();
 
+	const promiseRejectionHandler = (e: PromiseRejectionEvent) => {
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -34,6 +34,19 @@ const AppRootBase = () => {
 	const searchStore = useSearchStore();
 	const tabsStore = useTabsStore();
 
+	const promiseRejectionHandler = (e: PromiseRejectionEvent) => {
```
rzvdv, line 37: There are couple problems with this code, useEffect should have empty list of dependencies and this function should have been inside useEffect since it's not used anywhere else.

But i don't think that's a good approach to fix it. I think error should be handled where it occurred. I guess it comes from `MessagesSSEChannel.getFetchedChunk` and we need to find a way and place to gracefully handle it's cancellation

PR #363, opened 2022-05-18
TH2-3709. Invalid date in different parts of an application

VitalyMashanov:
* fixed

### DIFF
--- a/src/helpers/date.ts
+++ b/src/helpers/date.ts
@@ -42,8 +42,10 @@ export function formatTime(time: string | number) {
 	return moment.utc(time).format(DateTimeMask.DATE_TIME_MASK);
 }
 
-export function timestampToNumber(timestamp: Timestamp): number {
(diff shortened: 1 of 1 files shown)


src/helpers/date.ts
```diff
@@ -42,8 +42,10 @@ export function formatTime(time: string | number) {
 	return moment.utc(time).format(DateTimeMask.DATE_TIME_MASK);
 }
 
-export function timestampToNumber(timestamp: Timestamp): number {
-	return Math.floor(timestamp.epochSecond * 1000 + timestamp.nano / 1_000_000);
+export function timestampToNumber(timestamp: Timestamp | string): number {
```
rzvdv, line 45: if this function was called with string typescript would have notified an error about it, so it is weird that we are changing this function param types without changing models or something else

VitalyMashanov:
closing pull request due to using wrong 6th version of backend on 5th version of ui

PR #371, opened 2022-06-07
[ReportViewerUI] 6.1 Remove Graph from the UI

ponkk513:


src/components/messages-view-configurator/MessagesViewConfigurator.tsx
```diff
@@ -18,6 +18,7 @@ import React, { useRef, useState } from 'react';
 import { useOutsideClickListener } from '../../hooks';
 import { ModalPortal } from '../util/Portal';
 import '../../styles/messages-view-configurator.scss';
+import '../../styles/messages.scss';
```
rzvdv, line 21: there are too many imports of messages.scss. i think we can delete all these imports and use import in MessagesWindow

rzvdv (changes requested):
It looks like GraphStore is also redundant, can we delete it aswell?

src/stores/SelectedStore.ts
```diff
@@ -70,27 +68,6 @@ export class SelectedStore {
 		return this.workspacesStore.activeWorkspace.messagesStore.hoveredMessage;
```
rzvdv, line 68: hoveredMessage and hoveredEvent are unused now and should be deleted

rzvdv (changes requested):

src/components/messages-view-configurator/MessagesViewConfigurator.tsx
```diff
@@ -32,11 +32,17 @@ const MessageViewConfigurator = ({ sessions }: Props) => {
...
 	const buttonRef = useRef<HTMLButtonElement>(null);
+	const imageRef = useRef<HTMLImageElement>(null);
 
 	useOutsideClickListener(modalRef, (e: MouseEvent) => {
 		const isFromAutocomplete = Boolean((e.target as HTMLElement).closest('.rules-autocomplete'));
 		const isFromSelect = Boolean((e.target as HTMLElement).closest('.rules-select-options-list'));
-		if (e.target !== buttonRef.current && !isFromAutocomplete && !isFromSelect) {
+		if (
+			e.target !== buttonRef.current &&
+			e.target !== imageRef.current &&
```
rzvdv, line 42: instead of having 2 refs and checking every you can replace it with rootRef.contains(e.target)

src/components/messages-view-configurator/MessagesViewConfigurator.tsx
```diff
@@ -50,22 +56,30 @@ const MessageViewConfigurator = ({ sessions }: Props) => {
 		mode === 'body-sort' ? 'active' : null,
 	);
 
+	const offsetTop = document
+		.querySelector('.messages-window-header__configurator-button')
```
rzvdv, line 60: why would you query an element if you already have a ref. use buttonRef to get getBoundingClientRect values

src/stores/events/EventsFilterStore.ts
```diff
@@ -121,11 +133,26 @@ export default class EventsFilterStore {
 		});
 	}
 
+	@action
+	public setInterval = (interval: number) => {
```
rzvdv, line 137: i think this function is never called

PR #372, opened 2022-06-07
[TH2-3787] New workspace is created on graph search input blur

osinkinroman:


src/components/graph/search/GraphSearch.tsx
```diff
@@ -83,7 +83,7 @@ function GraphSearch(props: Props) {
 					inputConfig.isValidDate &&
 					inputConfig.timestamp <= moment.utc().valueOf()
 				) {
-					onTimestampSubmit(inputConfig.timestamp);
+					onTimestampSubmit(inputConfig.timestamp, true);
```
rzvdv, line 86: well with the same success we could have deleted this if block with function because now it never gets called, right?. But I'm confused about why did we even had this and what is expected behavior

rzvdv (approved):

PR #376, opened 2022-06-09
[TH2-3801] Graph Search: when searching for events, loading indicator doesn't appear

osinkinroman:


src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -56,7 +56,7 @@ const GraphSearchDialog = (props: Props) => {
 
 	const rootStore = useRootStore();
 
-	const [isLoading, setIsLoading] = React.useState(false);
+	const [isLoading, setIsLoading] = React.useState(true);
```
rzvdv, line 59: why is it true by default? it should be true only when search is active

rzvdv (changes requested):

rzvdv (approved):

PR #377, opened 2022-06-14
[TH2-3814] A message with a higher timestamp turned out to be lower than a message with a lower timestamp

osinkinroman:


rzvdv (changes requested):
i don't understand what message index or nextMessages.length has to do with incorrect sorting
Looks like it's a bug on report-data-provider

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -269,7 +270,9 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 				newMessagesList = newMessagesList.slice(-this.messagesLimit);
 			}
 
-			this.messages = newMessagesList;
+			this.messages = newMessagesList.sort(
```
rzvdv, line 273: we don't have to sort previous, it should be sorted already

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -170,6 +174,8 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 
 		if (!messageIds) return;
 
+		if (message) this.initialMessage = message;
```
rzvdv, line 177: we don't need this value. you should check messages[0]

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -295,10 +300,22 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
 			this.startIndex -= messages.length;
 
-			let newMessagesList = [...messages, ...this.messages];
-
+			let newMessagesList = [
+				...messages.filter(
+					message =>
+						!this.messages.includes(message) &&
+						message.messageId !== this.initialMessage?.messageId,
+				),
```
rzvdv, line 316: includes will always return false since objects references are different

osinkinroman: about it, I've just checked without it and it is actually working as it must work

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -295,10 +296,20 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		}
 
 		if (messages.length !== 0) {
+			const previousMessages = messages.filter(
+				message =>
+					timestampToNumber(message.timestamp) < timestampToNumber(this.messages[0].timestamp),
+			);
+			if (previousMessages.length !== 0) this.messages = [...this.messages, ...previousMessages];
```
rzvdv, line 303: what if this.messages.length is 1000 and previous messages should go after first message.
in the list would look like [firstMessages, 999 messages, previousMessages]

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -295,10 +296,20 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		}
 
 		if (messages.length !== 0) {
+			const previousMessages = messages.filter(
```
rzvdv, line 299: i think we need something like this
```
const prevMessages = getPrevMessages(); // will get messages with lesser timestamp or target
const firstNextMessage = messages[messages.length - 1];

if (firstNextMessage && firstNextMessage.messageId === this.messages[0]?.messageId) {
	prevMessages.pop();
}

const nextMessages = prevMessages.length ? messages.slice(0, prevMessages.length) : messages;

let newMessagesList = prevMessages.length
	? [
			...nextMessages,
			...[this.messages[0], ...prevMessages].filter(isEventMessage),
			...this.messages.slice(1),
			]
	: [...nextMessages, ...this.messages];

```

osinkinroman: ...[this.messages[0], ...prevMessages].filter(isEventMessage) this part I will change to ...prevMessages.filter(isEventMessage) because otherwise there is a copy of initial message[0] appears in the list

osinkinroman: and getPrevMessages I decided to not create as new private function and leave here, because there is a function with similar name getPreviousMessages

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -288,17 +289,27 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
+			prevMessages.pop();
 		}
 
 		if (messages.length !== 0) {
 			this.startIndex -= messages.length;
 
-			let newMessagesList = [...messages, ...this.messages];
-
+			let newMessagesList = prevMessages.length
+				? [...nextMessages, ...prevMessages.filter(isEventMessage), ...this.messages.slice(1)]
```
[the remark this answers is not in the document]
osinkinroman: image
this is what right now, and result is 
image

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -288,17 +289,27 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 	@action
 	public onNextChannelResponse = (messages: EventMessage[]) => {
 		this.lastNextChannelResponseTimestamp = null;
+
+		const prevMessages = messages.filter(
+			message =>
+				timestampToNumber(message.timestamp) <= timestampToNumber(this.messages[0].timestamp),
```
rzvdv, line 295: this should only check for less timestamp or id equal to first message.

rzvdv (approved):

PR #384, opened 2022-06-24
[TH2-3862] and [TH2-3863] 6.1 Implement message api fixes from 6.0 and add button for Search results to open in a new workspace

osinkinroman:


src/api/sse.ts
```diff
@@ -77,6 +77,17 @@ export interface EventSSEParams extends BaseSSEParams {
...
 
+export type MessageDirection = 'first' | 'second';
+
+export const toStream = (
+	name: string,
+	direction: MessageDirection[] = ['first', 'second'],
+): string[] => {
+	const res = direction.map(dir => `${name}:${dir}`);
+
+	return res;
```
rzvdv, line 88: ```return direction.map(dir => `${name}:${dir}`)```

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -15,19 +15,40 @@
...
+	const [messageList, setMessageList] = React.useState<EventMessageItem[]>([]);
 	const [viewType, setViewType] = useState(MessageViewType.JSON);
 	const [errorStatus, setErrorStatus] = useState<string | null>(null);
 
 	useEffect(() => {
 		getMessage();
 	}, []);
 
+	useEffect(() => {
+		if (message)
```
rzvdv, line 33: let's use braces {}

Does that code even work ? it looks like it should cause infinite loop. Effect has no dependencies and you changing state inside

src/components/embedded/EmbeddedEventHeader.tsx
```diff
@@ -31,6 +31,7 @@ export default function EmbeddedEventHeader(props: Props) {
 	const { eventId, eventName, eventType, startTimestamp, endTimestamp } = event;
 
 	const status = getEventStatus(event);
+	const startTimestampValue = timestampToNumber(startTimestamp);
```
rzvdv, line 34: lets change formatTime so it accepts string instead of importing timestampToNumber everywhere

src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -91,6 +101,29 @@ const MessagesVirtualizedList = (props: Props) => {
 		firstNextChunkIsLoaded,
 	]);
 
+	React.useEffect(() => {
+		toJS(messages).forEach(message =>
```
rzvdv, line 105: why do you need do convert it toJS ? we have shallow observable, so messages inside should be plain objects

src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -91,6 +101,29 @@ const MessagesVirtualizedList = (props: Props) => {
...
+			message.parsedMessages
+				? message.parsedMessages.forEach((parsedMessage, index) => {
```
rzvdv, line 107: i don't think using ternary here is a good idea

src/components/embedded/EmbeddedMessages.tsx
```diff
@@ -158,6 +150,31 @@ const MessagesVirtualizedList = observer((props: Props) => {
...
+				};
+
+				tempMessageItem.parsedMessage = tempMessage.parsedMessages
+					? tempMessage.parsedMessages[index]
+					: null;
+
+				if (tempMessageItem.parsedMessages && tempMessageItem.parsedMessage)
+					tempMessageItem.parsedMessages[0] = tempMessageItem.parsedMessage;
+
+				setMessageList([...messageList, tempMessageItem]);
```
rzvdv, line 173: so if we have 100 messages, and every message has 10 parsedMessages we call setMessageList 1000 times, right?

src/components/embedded/EmbeddedMessages.tsx
```diff
@@ -158,6 +150,31 @@ const MessagesVirtualizedList = observer((props: Props) => {
 		scrolledIndex,
 	} = props;
 
+	const [messageList, setMessageList] = React.useState<EventMessageItem[]>([]);
+
+	React.useEffect(() => {
+		messagesStore.dataStore.messages.forEach(message =>
+			message.parsedMessages?.forEach((parsedMessage, index) => {
+				const tempMessage = message;
```
rzvdv, line 158: is this variable even needed?

src/components/embedded/EmbeddedMessages.tsx
```diff
@@ -158,6 +150,31 @@ const MessagesVirtualizedList = observer((props: Props) => {
 		scrolledIndex,
 	} = props;
 
+	const [messageList, setMessageList] = React.useState<EventMessageItem[]>([]);
+
+	React.useEffect(() => {
+		messagesStore.dataStore.messages.forEach(message =>
+			message.parsedMessages?.forEach((parsedMessage, index) => {
```
rzvdv, line 157: it looks weird that parsedMessage is unused but index is used instead

rzvdv (changes requested):

src/api/sse.ts
```diff
@@ -186,6 +201,7 @@ export function getMessagesSSEParamsFromFilter(
 
 const sseApi: SSESchema = {
 	getEventSource: config => {
+		// console.error(new Error());
```
rzvdv, line 204: can we delete this?

src/components/message/message-card-list/MessagesVirtualizedList.tsx
osinkinroman: It's strange, but when I tested before it just didn't work right, so I used toJS, ok I removed it

PR #386, opened 2022-06-27
Version 6

rzvdv:


src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -41,15 +62,20 @@ function EmbeddedMessage({ messageId }: { messageId: string }) {
 		throw new Error(errorStatus);
 	}
 
-	if (message) {
+	if (messageList) {
```
rzvdv, line 65: messageList is an array, it will always be a truthy value

src/components/embedded/embedded-stores/EmbeddedMessagesDataProviderStore.tsx
```diff
@@ -187,10 +187,14 @@ export default class EmbeddedMessagesDataProviderStore implements MessagesDataSt
 
 		const [nextMessages, prevMessages] = await Promise.all([
 			this.searchChannelNext.loadAndSubscribe({
-				resumeMessageIds: extractMessageIds(messageIds.previous),
+				resumeMessageIds: extractMessageIds(
+					messageIds.next.filter((messageId, index) => index < 2),
```
rzvdv, line 191: this is probably wrong

src/components/graph/search/GraphSearchDialog.tsx
```diff
@@ -281,15 +283,34 @@ const GraphSearchDialog = (props: Props) => {
 				.getMessage(id, abortController.signal, { probe: true })
 				.then(foundMessage => {
 					if (foundMessage !== null) {
-						onSearchSuccess(foundMessage, searchTimestamp);
+						foundMessage.parsedMessages?.forEach((parsedMessage, index) => {
```
rzvdv, line 286: this code repeats at least 3 times, maybe we should extract it to a helper function?

src/helpers/date.ts
```diff
@@ -112,7 +112,7 @@ export function getRangeFromTimestamp(timestamp: number, interval: number): Time
 	];
 }
 
-export function sortByTimestamp<T extends { timestamp: number | Timestamp }>(
+export function sortByTimestamp<T extends { timestamp: number | string }>(
```
rzvdv, line 115: you can sort timestamp by subtracting a string

PR #391, opened 2022-07-19
[TH2-3967] Messages horizontal scroll activates downloading of new messages

osinkinroman:


src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -121,7 +128,10 @@ const MessagesVirtualizedList = (props: Props) => {
 					searchChannelPrev &&
 					!searchChannelPrev.isLoading &&
 					!searchChannelPrev.isEndReached &&
-					(wheelScrollDirection === undefined || wheelScrollDirection === 'previous')
+					!isHorizontal &&
+					((wheelScrollDirection === undefined &&
+						scroller.scrollHeight !== scroller.clientHeight) ||
```
rzvdv, line 133: why do we need this line ? i think this can lead to a bug when messages will not load.
let's say there no next messages, and all previous messages fitted into screen without overflowing it. in this case scrollHeight === clientHeight and you wont be able to load more previous messages

osinkinroman: This is for the case with horizontal scroll without wheel, I'll look how to change it

src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -101,7 +101,11 @@ const MessagesVirtualizedList = (props: Props) => {
 	]);
 
 	const debouncedScrollHandler = useDebouncedCallback(
-		(event: React.UIEvent<'div'>, wheelScrollDirection?: 'next' | 'previous') => {
+		(
```
rzvdv, line 104: i wonder why does scroll event event triggers when message is scrolled.
can we try rewriting it completely

There is a scrollerRef in Virtuoso

I think if rewrite it this way it would be easier to understand what's happening

```
useEffect(() => {
	function onScroll() {}
	function onWheel() {}

	scrollerRef.addEventListener('scroll', onScroll)
	scrollerRef.addEventListener('wheel', onWheel)
} ,[loadNext, loadPrev])

```

rzvdv (approved):

PR #402, opened 2022-07-29
[TH2-4020] Add collapsed view for messages

osinkinroman:


package.json
```diff
@@ -16,6 +16,7 @@
 		"licenses": "license-checker --production --csv --out licenses.csv"
 	},
 	"dependencies": {
+		"@react-hook/resize-observer": "^1.2.6",
```
rzvdv, line 19: can we implement it without adding enw dependency? we already have resize-observer-polyfill and use it in project

src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -64,6 +65,14 @@ const MessageCardBase = React.memo(
 	}: MessageCardBaseProps) => {
 		const { id, rawMessageBase64 } = message;
 
+		const wrapperRef = React.useRef(null);
+
+		const wrapperWidth = useElementSize(wrapperRef)?.width;
```
rzvdv, line 70: Wouldnt measuring list element more right way to do it?

src/components/message/message-card/header/MessageCardHeader.tsx
```diff
@@ -96,23 +98,27 @@ export const MessageCardHeader = React.memo((props: MessageInfoProps & MessageCa
...
 				{isAttached && <div className='mc-header__attached-icon' />}
 			</Chip>
-			<Chip
-				className={timestampClassName}
-				title={`Timestamp: ${formattedTimestamp}`}
-				onMouseEnter={onTimestampMouseEnter}
-				onMouseLeave={onTimestampMouseLeave}>
-				{timestamp && formattedTimestamp}
-			</Chip>
+			{!isCollapsed && (
```
rzvdv, line 101: Instead of boolean can we have an enum just as we use in EventCardHeader?
```
const enum CardDisplayType {
	STATUS_ONLY = 'status-only',
	MINIMAL = 'minimal',
	FULL = 'full',
}
```

rzvdv (changes requested):

src/components/message/message-card/MessageCardBase.tsx
```diff
@@ -57,13 +59,28 @@ const MessageCardBase = React.memo(
...
 		toogleMessagePin,
+		isEmbedded,
 		sortOrderItems,
 		addMessageToExport,
 		isExport,
 		isExported,
 	}: MessageCardBaseProps) => {
 		const { id, rawMessageBase64 } = message;
 
+		const messagesListRef = isEmbedded
```
rzvdv, line 70: use Virtuoso scrollerRef prop. and observe it's width

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -27,6 +29,13 @@ function EmbeddedMessage({ messageId }: { messageId: string }) {
 	const [rawViewType, setRawViewType] = useState(MessageViewType.ASCII);
 	const [isExpanded, setIsExpanded] = React.useState(true);
 
+	const wrapperWidth = useElementSize(document.body as HTMLDivElement)?.width;
```
rzvdv, line 32: body isnt HTMLDivElement. useElementSize should paramater should be of type HTMLElement.
and lets use ref for this one aswell. querying document for elements isnt react way to do it

src/components/embedded/EmbeddedMessages.tsx
```diff
@@ -248,11 +256,23 @@ const MessagesVirtualizedList = observer((props: Props) => {
 		debouncedScrollHandler(event, event.deltaY < 0 ? 'next' : 'previous');
 	};
 
+	const itemRenderer = React.useCallback(
```
rzvdv, line 259: i think its still better to accept itemRenderer as a prop.
you can do it by creating another another function like so

```
itemContent={(index, message) => {
  return itemRenderer(message, displayType)
}}
```

also we can omit index parameter in itemRenderer since it isnt very usefull

osinkinroman: index is required by itemContent prop of Virtuoso, so I think we can't delete it

src/hooks/useElementSize.ts
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************** */
+import * as React from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
+
+const useElementSize = (target: HTMLElement | null) => {
+	console.log(target);
```
rzvdv, line 20: remove console.log

src/hooks/useElementSize.ts
```diff
@@ -0,0 +1,40 @@
...
+	const [size, setSize] = React.useState<DOMRectReadOnly>();
+
+	const resizeObserver = new ResizeObserver(entries =>
```
rzvdv, line 23: this will create observer each time, can we store it in ref&

src/components/embedded/EmbeddedMessage.tsx
```diff
@@ -19,13 +19,26 @@ import { EventMessage, MessageViewType, MessageViewTypeConfig } from '../../mode
...
+import useElementSize from '../../hooks/useElementSize';
+import CardDisplayType, { COLLAPSED_MESSAGES_WIDTH } from '../../util/CardDisplayType';
 
 function EmbeddedMessage({ messageId }: { messageId: string }) {
 	const [message, setMessage] = useState<EventMessage | null>();
 	const [errorStatus, setErrorStatus] = useState<string | null>(null);
 	const [viewType, setViewType] = useState(MessageViewType.JSON);
 	const [rawViewType, setRawViewType] = useState(MessageViewType.ASCII);
 	const [isExpanded, setIsExpanded] = React.useState(true);
+	const [wrapperContent, setWrapperContent] = React.useState<HTMLDivElement | null>(null);
```
rzvdv, line 31: there is  a useRef hook specifically for refs

rzvdv (dismissed):

PR #409, opened 2022-08-12
[TH2-4080] New filters panel

osinkinroman:


src/components/event/EventWindow.tsx
```diff
@@ -45,6 +46,9 @@ function EventWindow() {
 					onSelect={onBreadcrumbItemClick}
 				/>
 			</div>
+			<div className='window__controls'>
```
rzvdv, line 49: so now its 2 nested div.window__controls

rzvdv (changes requested):
so there is now no EventsFilterPanel component, but there is a MessageFilterPanel component which isn't really a filter panel but just a session input.

can we have EventsFilterPanel and MessageFilterPanel components back?

And the way filter panel opens is weird, either it has to have absolute positioning or list should persist it's height.
Because now after it opens it shrinks the list down like to 3 rows

resources/icons/or.svg
```diff
@@ -0,0 +1,18 @@
+<svg width="28" height="29" viewBox="0 0 28 29" fill="none" xmlns="http://www.w3.org/2000/svg">
```
rzvdv, line 1: so & is OR ?

src/styles/messages.scss
```diff
@@ -878,6 +878,7 @@
 	height: 100%;
 	width: 100%;
 	display: flex;
+	flex-direction: column;
```
rzvdv, line 881: it has only 1 div as a child, does ir really need flex-direction:column? i think some other css is redundant as well

src/components/message/ReplayModal.tsx
```diff
@@ -194,7 +194,11 @@ function ReplayModal() {
 					<button className='replay__close-button' onClick={() => setIsOpen(false)}>
 						<i></i>
 					</button>
-					<FilterRows config={filterConfig} />
+					<FilterRows
+						config={filterConfig}
+						headerClassName={'replay__compound-header'}
+						onMouseOver={() => setDrag(false)}
```
rzvdv, line 200: what does this line do? i dont get it

src/components/search-panel/SearchPanelFilters.tsx
```diff
@@ -20,9 +20,7 @@ import {
 	FilterRowTogglerConfig,
 	FilterRowSwitcherConfig,
 } from '../../models/filter/FilterInputs';
-import { SSEFilterInfo } from '../../api/sse';
-import { SearchPanelType } from './SearchPanel';
-import { FiltersHistoryType } from '../../stores/FiltersHistoryStore';
+import '../../styles/filter.scss';
```
rzvdv, line 23: this import is kind of implicit. why does this file import styles if it doesnt have any components?
and i believe removed this file in 6.1

src/components/filter/row/TogglerRow.tsx
```diff
@@ -19,27 +19,29 @@ import { FilterRowTogglerConfig } from '../../../models/filter/FilterInputs';
...
 
 const TogglerRow = ({ config }: { config: FilterRowTogglerConfig }) => {
-	const { value, toggleValue, possibleValues, label, disabled, className = '' } = config;
-	const [firstLabel, secondLabel] = possibleValues;
+	const { value, toggleValue, possibleValues, disabled } = config;
+	const [firstPossibleValue, secondPossibleValue] = possibleValues;
 
-	const togglerClassName = createStyleSelector('toggler', disabled ? 'disabled' : '');
-	const togglerBarClassName = createStyleSelector('toggler__bar', className, value ? 'on' : 'off');
+	const firstLabel = firstPossibleValue === 'excl' ? 'Exclude' : 'And';
```
rzvdv, line 25: can these properties be renamed in useFilterConfig.togglerValues ?

resources/icons/or.svg
osinkinroman: So I've asked about it Anatoly, about maybe it will be better to make additional icon for or, he said that's leave as it is right now, grey icon for 'or' and blue for 'and'

src/components/filter/row/SwitcherRow.tsx
```diff
@@ -26,15 +26,10 @@ const SwitcherRow = ({ config }: { config: FilterRowSwitcherConfig }) => {
 		}
 	};
 
-	const labelClassName = createStyleSelector('filter-row__label', config.labelClassName || null);
+	const rootClassName = config.labelClassName ? 'filter__compound-header' : 'search-type-switcher';
```
rzvdv, line 29: it doesnt have label now, but it accepts labelClassName?

src/components/filter/row/MultipleStringFIlterRow.tsx
```diff
@@ -93,6 +93,7 @@ function MultipleStringFilterRow({ config }: MultipleStringFilterRowProps) {
 		'multiple-values',
 		'filter-row__input',
 		config.isInvalid ? 'invalid' : null,
+		isFocused ? 'active' : null,
```
rzvdv, line 96: overflow is broken

rzvdv (approved):

PR #414, opened 2022-08-17
Loading target event

atachka:


src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -25,13 +27,35 @@ export class ExperimentalAPIEventStore {
 
 	@observable scrolledIndex: null | Number = null;
 
-	constructor() {
+	events: EventStoreDefaultStateType;
+
+	constructor(events: EventStoreDefaultStateType) {
```
rzvdv, line 32: you dont neet to store it since we only need in constructor

rzvdv, line 32: can you rename it to initialState? it would be more clear what it is

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -25,13 +27,35 @@ export class ExperimentalAPIEventStore {
...
+		this.events = events;
 		this.getRootIds();
+		this.loadEvent((events as any)?.selectedEventId);
```
rzvdv, line 35: try doing it without using any.
first check if it's a string then it will be an id we need. if it's not a string check if its not a null

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -25,13 +27,35 @@ export class ExperimentalAPIEventStore {
...
 	}
 
+	@action
+	loadEvent = async (eventId: string) => {
+		if (eventId) {
+			const data = await this.fetchEvent(eventId);
+			const firstChild = JSON.parse(JSON.stringify(data));
```
rzvdv, line 42: doesnt look like this line doing anything. are you trying to create a copy of data?

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -25,13 +27,35 @@ export class ExperimentalAPIEventStore {
...
+			const parents = await eventHttpApi.getEventParents(firstChild.eventId);
```
rzvdv, line 43: look's like `getEventParents` requires firstParentId as an argument. which probably a bit confusing. so you don't even have to fetch first event separately with `this.fetchEvent(eventId)`,

getEventParents will return events up to root with eventId included

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -25,13 +27,35 @@ export class ExperimentalAPIEventStore {
...
+			for (let i = 0; i < parents.length; i++) {
+				this.cache.set(parents[i].eventId, parents[i]);
+				if (parents[i].parentEventId) {
+					this.parentsChildrenMapMap.set(parents[i].parentEventId, [parents[i].eventId]);
+				}
+				this.isExpanded.set(parents[i].parentEventId, true);
+			}
+			this.scrollToEvent(eventId);
```
rzvdv, line 51: you also need to select target event here

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -119,11 +143,14 @@ export class ExperimentalAPIEventStore {
 	};
 
 	getParentNodes(eventId: string, cache: Map<string, EventAction>): EventAction[] {
-		let event = cache.get(eventId);
+		const event = cache.get(eventId);
```
rzvdv, line 146: i think this function should not have been modified

rzvdv (changes requested):

rzvdv (approved):

PR #416, opened 2022-08-17
add events pagination with offset to server

tesohi:


src/components/event/experimental-api/EventTree.tsx
```diff
@@ -53,6 +53,7 @@ function EventTree({ eventId }: EventTreeProps) {
 
 	const onExpandClick = () => {
 		eventsStore.toggleExpand(eventId);
+		eventsStore.clearEventChildren(eventId);
 		eventsStore.fetchChildren(eventId);
```
rzvdv, line 57: Now since we have pagination i think `eventsStore.fetchChildren` should be called here only when it's expanded for the first time

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -130,6 +140,16 @@ export class ExperimentalAPIEventStore {
 		this.isExpanded.set(eventId, !this.isExpanded.get(eventId));
 	};
 
+	@action clearEventChildren(eventId: string) {
+		if (eventId === undefined) return;
```
rzvdv, line 144: I don't think we even need this function. we definetely dont want to reset parentsChildrenMapMap. btw can you rename it and remove extra Map word?

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -109,12 +113,18 @@ export class ExperimentalAPIEventStore {
 		if (this.isLoadingChildren.get(eventId) || this.hasMoreChildren.get(eventId) === false) return;
 		const currentChildren = this.parentsChildrenMapMap.get(eventId) || [];
 		try {
-			const ids = await api.events.getChildrenIds({ parentId: eventId });
+			const ids = await api.events.getChildrenIds({
+				parentId: eventId,
+				limit: this.CHILDREN_COUNT_LIMIT + 1,
+				offset: (this.parentsChildrenChunks.get(eventId) || 0) * this.CHILDREN_COUNT_LIMIT,
```
rzvdv, line 119: I think we can just use `parentsChildrenMapMap.get(id).length || 0` and remove this extra state

src/components/event/experimental-api/ExperimentalAPIEventStore.ts
```diff
@@ -109,12 +113,18 @@ export class ExperimentalAPIEventStore {
 		if (this.isLoadingChildren.get(eventId) || this.hasMoreChildren.get(eventId) === false) return;
 		const currentChildren = this.parentsChildrenMapMap.get(eventId) || [];
 		try {
-			const ids = await api.events.getChildrenIds({ parentId: eventId });
+			const ids = await api.events.getChildrenIds({
```
rzvdv, line 116: also need to add `this.isLoadingChildren(id, true)`

rzvdv (changes requested):

rzvdv (approved):

PR #418, opened 2022-08-23
fixed the filtration

atachka:


rzvdv (changes requested):
When there is a JIRA ticket you should name your commit and pr like so https://github.com/th2-net/th2-rpt-viewer/pull/395/commits

You shouldn't be using casting.  We have type guard functions for this. Search for isEvent, isEventMessage

Don't commit commented code

You need to create a helper function, something like getBookmarkTitle in helpers/bookmarks
And this function should be used in BookmarkItemBase component to render a title and in BookmarksStore to search for a query in the title.

Usually you would also need to bump a patch version in package.json in this case it would be 6.0.12

src/stores/BookmarksStore.ts
```diff
@@ -73,6 +73,7 @@ export class BookmarksStore {
 	@computed
 	public get filteredBookmarks() {
 		const search = this.textSearch.toLowerCase();
+		console.log(this.sortedBookmarks);
```
rzvdv, line 76: Don't commit commented code

Usually you would also need to bump a patch version in package.json in this case it would be 6.0.14

src/helpers/event.ts
```diff
@@ -112,8 +112,14 @@ export function getItemId(item: EventAction | EventTreeNode | EventMessage) {
...
 
 export function getItemName(item: EventAction | EventTreeNode | EventMessage) {
-	if (isEventMessage(item)) return item.parsedMessages?.[0].message.metadata.messageType;
-	return item.eventName;
+	if (isEventMessage(item) && item.parsedMessages) {
+		return item.parsedMessages.reduce((previous, current) => {
+			return previous + current.message.metadata.messageType.concat(',');
+		}, '');
+	}
+	if (isEventAction(item)) return item.eventName;
```
rzvdv, line 120: There is another problem here, it doesnt handle EventTreeNode

add isEventNode type guard. it should return eventName as well

rzvdv (approved):

PR #423, opened 2022-08-30
[Th2-4021] added check to messages when soft filter

molotgor:


src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -200,7 +200,13 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		}
 
 		if (this.messagesStore.filterStore.isSoftFilter && message) {
```
rzvdv, line 202: can we just delete this if-block, it seems to be redundant, and it should fix the bug

molotgor: We need to keep if-block for message otherwise it will shout that object is possibly null. And do we need to match message when we don't have softFilter?

rzvdv: i mean delete this if-block entirely. when soft filter is on MessageCard has a hook to make a request for matching filter. This is weird optimization that i made by mistakenly expecting that selectedMessageId will always match the filter

rzvdv (approved):

PR #430, opened 2022-09-05
[TH2-4134] Message, linked to event always remain in the list, regardless of the applied filters

osinkinroman:


src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -190,7 +190,15 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 				...nextMessages.filter(val => val.messageId !== message?.messageId),
 				...[message].filter(isEventMessage),
 				...prevMessages,
-			].sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp));
+			]
+				.filter(
+					item =>
+						!(
+							this.messagesStore.filterStore.filterParams['attachedEventIds-negative'] &&
```
rzvdv, line 197: are you sure it's reproduced only when attachedEventIds filter applied?

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -498,4 +503,22 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 			});
 		}
 	};
+
+	@action
+	private isMessageExcluded = (message: EventMessage) => {
```
rzvdv, line 508: There is an api to check if message matches filter

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -268,7 +269,9 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		}
 
 		if (messages.length) {
-			let newMessagesList = [...this.messages, ...messages];
+			let newMessagesList = [...this.messages, ...messages].filter(
+				val => !this.isMessageExcluded(val),
```
rzvdv, line 273: I don't think you need to check it here

rzvdv (changes requested):

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -143,11 +143,19 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
 		if (this.messagesStore.selectedMessageId) {
 			try {
-				this.messageAC = new AbortController();
-				message = await this.api.messages.getMessage(
-					this.messagesStore.selectedMessageId.valueOf(),
-					this.messageAC.signal,
-				);
+				if (
+					await this.api.messages.matchMessage(
+						this.messagesStore.selectedMessageId.valueOf(),
```
rzvdv, line 148: this.messagesStore.selectedMessageId is still used to fetch messageIds even if doesnt match the filter which is probably wrong.
so i guess we need to reset it here

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -143,11 +143,19 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
+						this.messagesStore.filterStore.filterParams,
+					)
+				) {
+					this.messageAC = new AbortController();
+					message = await this.api.messages.getMessage(
+						this.messagesStore.selectedMessageId.valueOf(),
+						this.messageAC.signal,
+					);
+					this.messagesStore.selectedMessageId = null;
```
rzvdv, line 157: i mean you should reset it if tt doesnt match the filter

PR #431, opened 2022-09-08
[TH2-4143] changed older/newer buttons in eventsPanelHeader to fetch older/newer…

molotgor:


src/modules/events/stores/EventsStore.ts
```diff
@@ -468,6 +469,56 @@ export default class EventsStore implements IEventsStore {
 		});
 	};
 
+	@observable searchChannel: EventSource | null = null;
+
+	public changeToClosestEvent = (
+		searchDirection: SearchDirection.Previous | SearchDirection.Next,
+	) => {
+		if (this.searchChannel) this.searchChannel.close();
```
rzvdv, line 477: - What if found event doesnt match current filter
- This looks weird. if timestampFrom = 12.00.00 and closest event is 11.59.59 we will just move 1 seconds backwards and reload whole tree from 11.59.59 ?

src/modules/events/stores/EventsStore.ts
```diff
@@ -468,6 +469,56 @@ export default class EventsStore implements IEventsStore {
...
+			type: 'event',
+			queryParams,
+		});
+		this.searchChannel = searchChannel;
+		searchChannel.addEventListener('event', this.changeEvent.bind(this, searchDirection));
+	};
+
+	public changeEvent = (direction: SearchDirection, event: MessageEvent) => {
+		const data = (event as MessageEvent).data;
+		const parsedEvent: EventTreeNode = JSON.parse(data);
```
rzvdv, line 498: i believe it might be SSEHeartbeat as well

src/modules/events/stores/EventsStore.ts
molotgor: If timestampFrom is 12.00.00 and closest event is 11.59.59 than we will move 1 second and we get new tree with 11.59.59 as timestampTo in range and (11.59.59 - interval) as timestampFrom

rzvdv (approved):

PR #464, opened 2022-10-31
[TH2-4389] Update button in messages doesn't work

osinkinroman:


src/stores/messages/MessagesUpdateStore.ts
```diff
@@ -39,14 +46,13 @@ export default class MessagesUpdateStore {
...
+		this.isFirstUpdate = true;
 		this.messagesDataStore.resetState();
 		this.messagesStore.selectedMessageId = null;
 		this.messagesStore.filterStore.filter.timestampTo = Date.now();
 
 		this.messagesDataStore.loadMessages({
-			onClose: async messages => {
-				this.messagesDataStore.onNextChannelResponse(messages);
-
+			onClose: async () => {
```
rzvdv, line 55: i believe the messages you ignore here wont appear in the list

src/stores/SSEChannel/MessagesSSEChannel.ts
```diff
@@ -153,7 +155,30 @@ export class MessagesSSEChannel extends SSEChannel<EventMessage> {
 	private _onMessageIdsEvent = (e: Event) => {
 		const messagesIdsEvent: MessageIdsEvent =
 			e instanceof MessageEvent && e.data ? JSON.parse(e.data) : null;
-		this.messageIds = Object.values(messagesIdsEvent.messageIds).filter(Boolean) as string[];
+		const streams = getArrayOfUniques(
+			this.messageIds
+				.filter(id => id.slice(id.lastIndexOf(':') + 1) === '-1')
```
rzvdv, line 160: `id.slice(-2) == '-1'` looks more readable i think?
btw if we extract it to helper functions it would be even more readable
`this.messageIds.filter(isInvalidResumeId).map(getSession)`

src/components/message/message-card-list/MessagesVirtualizedList.tsx
```diff
@@ -147,6 +147,11 @@ const MessagesVirtualizedList = (props: Props) => {
 
 	const onWheel: React.WheelEventHandler<'div'> = event => {
 		event.persist();
+
+		if (updateStore.isActive) {
+			updateStore.isActive = false;
```
rzvdv, line 152: I think we disscused this issue, and i dont remember what we've decided, but it looks unintuitive that it stops only on mouse wheel scroll

src/stores/SSEChannel/MessagesSSEChannel.ts
```diff
@@ -102,7 +105,7 @@ export class MessagesSSEChannel extends SSEChannel<EventMessage> {
 						this.getFetchedChunk(),
 				  ])
 				: this.getFetchedChunk());
-			return messagesChunk;
+			return messagesChunk.map(val => toJS(val)).slice();
```
rzvdv, line 108: messagesChunk isn't observable

src/components/message/message-card-list/MessagesVirtualizedList.tsx
osinkinroman: We've decided that if there is no way to also check regular scroll, then we don't need to implement this feature at all

rzvdv (approved):

PR #474, opened 2022-11-15
[TH2-4402] Timeline navigation arrows do not appear on the timeline a user to move gradually on the timeline

okti-web:


src/components/graph/OutsideItems.tsx
```diff
@@ -86,11 +105,108 @@ const OutsideItems = (props: OverlayPanelProps) => {
...
-		const panel = panels.sort((p1, p2) => p1.range[0] - p2.range[1])[0];
 
-		if (panel) {
-			onPanelRangeSelect(panel.range);
+		const windowRange = to - from;
+		const windowCenter = from + windowRange / 2;
+
+		// the first condition is needed in case  t > 30min.
+		// then there is no point in centering while inside the event.
+		const center0 =
```
rzvdv, line 116: this code is really hard to follow. it has too many if statements and repeating code.
it's possible i think to have only 1 if-statement
and this if statement should check if panel range is bigger than window range

repeating functions and constants should be predefined

5000 could be a padding

`panels[0].range[1] > from && panels[0].range[0] < from`
could be a function isIntersected which check if window border is inside panel range

there are a lot of comments which i don't find very useful. if pr is a draft it's ok. but if it's ready for review please remove all unnecessary ones.

i made a pr to speed things up. please check if there any corrections need to be made, and if it's okay merge it
https://github.com/th2-net/th2-rpt-viewer/pull/479/files

rzvdv (changes requested):

rzvdv (approved):

PR #484, opened 2022-11-24
Version 6.1 time selector

okti-web:


src/modules/events/components/EventListNavigation.tsx
```diff
@@ -34,13 +35,20 @@ export const EventListNavigation = observer(() => {
 
 	return (
 		<div className='events-nav'>
-			<button className='button-base' onClick={getPrevEvents}>
+			<div className='events-nav__arrow-wrapper events-nav__arrow-wrapper--left'  onClick={getPrevEvents}>
+				<div className='events-nav__arrow-icon' />
```
rzvdv, line 39: since 6.1 we started using icons as component. refer to /components/icons
dont use svg as background

PR #485, opened 2022-11-24
6.1 Events navigation buttons

okti-web:


src/modules/events/components/EventListNavigation.tsx
```diff
@@ -34,13 +38,22 @@ export const EventListNavigation = observer(() => {
...
+				className='events-nav__arrow-wrapper events-nav__arrow-wrapper--left'
+				onClick={getPrevEvents}>
+				<EventsArrowIcon />
+			</Chip>
+			<button className='button-base button-base--events' onClick={getPrevEvents}>
 				Show previous
 			</button>
 			<span className='events-nav__timestamp'>{formatTimestamp(timestamp)}</span>
-			<button className='button-base' onClick={getNextEvents}>
+			<button className='button-base button-base--events' onClick={getNextEvents}>
```
rzvdv, line 50: I think it's better to move chip inside the button instead of having 2 html elements with onClick handler, one of which isnt button

src/modules/events/components/EventListNavigation.tsx
```diff
@@ -34,13 +38,22 @@ export const EventListNavigation = observer(() => {
 
 	return (
 		<div className='events-nav'>
-			<button className='button-base' onClick={getPrevEvents}>
+			<Chip
+				className='events-nav__arrow-wrapper events-nav__arrow-wrapper--left'
```
rzvdv, line 42: we actually don't use modifiers like that. this way you should repeat whole class name twice
instead just declare a class .left
and classname will be `events-nav__arrow-wrapper left`


also wrapper is a toxic word for css classes i think, once you start doing that everything becomes a wrapper

src/styles/theme.scss
```diff
@@ -83,6 +85,10 @@ html[data-theme='dark'] {
 	--panel-title: #2d3036;
 
 	// rest
+	--info-color: #71a2ff;
```
rzvdv, line 88: i dont think that is true, there are only a few places where this hex color is used
it might be an error

src/styles/theme.scss
```diff
@@ -83,6 +85,10 @@ html[data-theme='dark'] {
...
+	--timestamp-color: #c6daff;
+	--arrow-background: #4f5156; 
+	--arrow-color: #a7a8ab;
```
rzvdv, line 91: this is too much for arrow to have a variable for it's color

src/modules/events/components/EventListNavigation.tsx
okti-web: should i change it to events-nav__arrow?

rzvdv: events-nav__arrow is ok

rzvdv (approved):

PR #490, opened 2022-11-28
6.1 search history

okti-web:


src/styles/filters-history.scss
```diff
@@ -133,24 +130,18 @@
...
 		position: relative;
-		& i {
-			position: absolute;
-			top: 50%;
-			left: 50%;
-			transform: translate(-50%, -50%);
-			opacity: 0.5;
-			@include icon(url(../../resources/icons/share-option.svg), 14px, 14px);
-		}
+		width: 14px;
```
okti-web, line 133: when inspecting the share-icon element, it shows a giant box 500x500. not sure how to fix it, and whether it is important.

rzvdv (changes requested):
Add styles to a button that opens this modal.
Also buttons on the item dont look too nice, can you convert them from svg to a components and add some margins, active styles, hover styles

src/styles/filter.scss
```diff
@@ -367,9 +367,9 @@
 	gap: 12px;
 	font-size: 14px;
 
-	.filters-history-open {
-		@include improved-clickable(#fff, 10px);
-	}
+	// .filters-history-open {
+	// 	@include improved-clickable(var(--background-primary), 10px);
+	// }
```
okti-web, line 372: not sure, whether this line is functional. getting rid of it doesn't change much for filter-history icon in Events.

osinkinroman: you can delete it, because currently history icon is not in filter-controls div

src/styles/theme.scss
```diff
@@ -61,6 +61,8 @@
 	--background-primary: #ffffff;
 	--background-secondary: #fbfbfb;
 
+	--background-primary-darker: hsl(0, 0%, 90%);
```
osinkinroman, line 64: I think it will be better to change colors like this to hex, because for the most of the code we use hex for color

src/styles/filter.scss
osinkinroman: delete it, cause you dont have class 'filter-controls filters-history-open'

src/styles/filters-history.scss
```diff
@@ -18,15 +18,30 @@
...
+	width: 40px;
+	height: 40px;
 	border: none;
+	border-radius: 12px;
+
+	display: flex;
+	justify-content: flex-start;
+	align-items: center;
+	padding-top: 10px;
+	margin-right: -5px;
```
osinkinroman, line 30: I dont think you need it, cause we have gap:12px in parent div, also try to use margin, padding and gaps in future to be multiple of 4

src/components/filters-history/FiltersHistoryItem.tsx
```diff
@@ -176,21 +180,21 @@ const FiltersHistoryItem = (props: Props) => {
 						onClick={onFilterPin}
 						ref={pinButtonRef}
 						title={item.isPinned ? 'Unpin filter' : 'Pin filter'}>
-						<i></i>
+						<PinIcon className='pin-icon' />
 					</button>
 					<button
-						className='filter-history-item__share-icon'
+						className='filter-history-item__filter-share-icon'
```
osinkinroman, line 186: I dont think that its necessary to rename it like that

osinkinroman (changes requested):

src/styles/filters-history.scss
```diff
@@ -117,40 +135,43 @@
...
-	&__delete-icon {
-		position: relative;
-		& i {
-			position: absolute;
-			transform: translate(-50%, -50%);
-			opacity: 0.5;
-			@include icon(url(../../resources/icons/cross-icon-dark.svg), 16px, 16px);
-		}
+	&__filter-delete-icon {
+		padding-right: 3px;
```
osinkinroman, line 169: if you need different paddings and margins, you can do it like padding: 1px 1px 1px 1px, and I also think that your approach is a bit difficult, try to use margin on parent __controls div

osinkinroman (changes requested):

PR #493, opened 2022-12-06
[TH2-4453] Implement new time picker

osinkinroman:


src/components/filter/date-time-inputs/TimeUnitList.tsx
```diff
@@ -71,24 +113,35 @@ const TimeUnitList = (props: TimeUnitListProps) => {
...
-					className={createBemElement(
-						'filter-timepicker',
-						'scroll-item',
-						selectedUnit === unitItem.value ? 'active' : null,
-						unitItem.isBlocked ? 'blocked' : null,
-					)}
-					onClick={!unitItem.isBlocked ? () => onUnitClick(unitItem.value) : undefined}>
-					{unitItem.value}
-				</li>
+				<>
```
molotgor, line 123: I think we don't need fragments here

molotgor (changes requested):
I think when we change date's hour/minute seconds should reset to 0

PR #495, opened 2022-12-12
[TH2-4479] messages are shuffled when History window is used for searching

osinkinroman:
session api-es4 used for test

### DIFF
--- a/src/stores/messages/MessagesDataProviderStore.ts
+++ b/src/stores/messages/MessagesDataProviderStore.ts
@@ -197,8 +197,23 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 
 		runInAction(() => {
 			const messages = [
-				...nextMessages.filter(val => val.messageId !== message?.messageId),
+				...nextMessages.filter(
+					(val, idx) =>
+						val.messageId !== message?.messageId &&
+						!(
+							message &&
+							timestampToNumber(val.timestamp) === timestampToNumber(message.timestamp) &&
+							idx === nextMessages.length - 1
+						),
+				),
 				...[message].filter(isEventMessage),
+				...nextMessages.filter(
+					(val, idx) =>
+						message &&
+						timestampToNumber(val.timestamp) === timestampToNumber(message.timestamp) &&
+						idx === nextMessages.length - 1 &&
+						val.messageId !== message?.messageId,
+				),
 				...prevMessages,
 			].sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp));
 			this.messages = messages;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "th2-rpt-viewer",
-	"version": "5.0.56",
+	"version": "5.0.57",
 	"description": "",
 	"main": "index.tsx",
 	"private": true,


osinkinroman:
I didn't find another way to resolve this problem

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -197,8 +198,15 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 
 		runInAction(() => {
 			const messages = [
-				...nextMessages.filter(val => val.messageId !== message?.messageId),
+				...nextMessages.filter(
```
rzvdv, line 201: so i guess the problem is the middle message that we put in center manually.
the way you can solve it here is just by getting unique messages, and than sorting them, i think that would work.

but you also have to solve the same problem in onNextChannelResponse
and i'm not sure how to do that

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -196,12 +195,9 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		]);
 
 		runInAction(() => {
-			const messages = [
-				...nextMessages.filter(val => val.messageId !== message?.messageId),
-				...[message].filter(isEventMessage),
-				...prevMessages,
-			].sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp));
-			this.messages = messages;
+			this.messages = Array.from(new Set([...nextMessages, ...prevMessages])).sort(
```
rzvdv, line 198: remove new Set

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -345,6 +339,7 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 
 		const nextMessages = messages
 			.slice(0, messages.length - prevMessages.length)
+			.filter(msg => !this.messages.map(msg => msg.messageId).includes(msg.messageId))
```
rzvdv, line 342: i thought onNextChannelResponse can be simplified a lot since we dont have to worry about order now
it should just insert new messages in the list without any extra checks

would it work like so?
```
public onNextChannelResponse = (nextMessages) => {
      this.lastNextChannelResponseTimestamp = null;
       this.messages = [...nextMessages, ...this.messages]
}

```

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -196,12 +195,9 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
 		]);
 
 		runInAction(() => {
-			const messages = [
-				...nextMessages.filter(val => val.messageId !== message?.messageId),
-				...[message].filter(isEventMessage),
-				...prevMessages,
-			].sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp));
-			this.messages = messages;
+			this.messages = [...nextMessages, ...prevMessages].sort(
```
rzvdv, line 198: i think we dont event have to sort at manually it should be sorted already

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -328,37 +321,22 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
-				? [...nextMessages, this.messages[0], ...prevMessages, ...this.messages.slice(1)]
-				: [...nextMessages, ...this.messages];
+		if (nextMessages.length) {
+			let newMessagesList = [...nextMessages, ...this.messages];
+
+			this.startIndex -= this.updateStore.isActive ? 0 : nextMessages.length;
 
 			if (newMessagesList.length > this.messagesLimit) {
-				newMessagesList = newMessagesList.slice(0, this.messagesLimit);
+				newMessagesList = newMessagesList.slice(-this.messagesLimit);
```
rzvdv, line 336: i think it's source of a bug and we should delete it.
because loaders keep last messageIds, and if you slice some messages loading will start with incorrect messageIds since there is no logic to refetch them

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -328,37 +321,22 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
-							message =>
-								timestampToNumber(message.timestamp) <
-								timestampToNumber(this.messages[0].timestamp),
-						)
-						.sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp))
-				: [];
-
 		const nextMessages = messages
-			.slice(0, messages.length - prevMessages.length)
+			.filter(message => !this.messages.map(msg => msg.messageId).includes(message.messageId))
```
rzvdv, line 325: instead of creating an array of ids for every message, you could have used .find method

if there is a problem about duplicating messages even a better way to do this would be to have a computed value Map<messageId, true> from messages list, so this check would be more efficient

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -328,37 +321,22 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
 			.sort((a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp));
```
rzvdv, line 326: i dont think sorting is necessary

osinkinroman: I also have a question about this line from master: 
			this.messages = [...new Map(newMessagesList.map(item => [item.messageId, item])).values()];
It looks weird, should I get it back or just newMessagesList is enough?

rzvdv: ye, im not sure if Map.values() guarantees same order
i guess it was made for filtering duplicates, so it could be filtered the same way you will filter it in onNextChannelResponse

src/stores/messages/MessagesDataProviderStore.ts
osinkinroman: So there is no duplicates problem as far as I see now, so I'll leave it with .find, if the bug will appear in future will use Map<messageId, true>

src/stores/messages/MessagesDataProviderStore.ts
```diff
@@ -328,37 +320,14 @@ export default class MessagesDataProviderStore implements MessagesDataStore {
...
-		if (prevMessages.length > 0 || nextMessages.length > 0) {
+		if (nextMessages.length) {
 			this.startIndex -= this.updateStore.isActive ? 0 : nextMessages.length;
 
-			let newMessagesList = prevMessages.length
-				? [...nextMessages, this.messages[0], ...prevMessages, ...this.messages.slice(1)]
-				: [...nextMessages, ...this.messages];
-
-			if (newMessagesList.length > this.messagesLimit) {
-				newMessagesList = newMessagesList.slice(0, this.messagesLimit);
```
rzvdv, line 359: delete this in previous handler as well

rzvdv (approved):

PR #496, opened 2022-12-12
[TH2-4448] [rpt-viewer][flat view] Sometimes the event is outside the visible area of the screen if the workspace is opened via sharing link (Get Workspace Link)

okti-web:


src/stores/events/EventsStore.ts
```diff
@@ -89,6 +89,13 @@ export default class EventsStore {
 		reaction(() => this.hoveredEvent, this.onHoveredEventChange);
 
 		reaction(() => this.graphStore.eventInterval, this.onIntervalChange);
+
+		reaction(
+			() => this.flattenedEventList,
+			() => {
+				if (this.selectedEvent) this.scrollToEvent(this.selectedEvent.eventId);
```
rzvdv, line 96: this will scroll to event every time event list is updated

rzvdv (dismissed):

osinkinroman (approved):

PR #505, opened 2023-01-09
[Th2-4580] added check when autocomplete is fully open

molotgor:


osinkinroman (commented):
What if we also speed up animation? in util.scss line 173, I guess It will be better if it faster render the animation

src/components/util/AutocompleteInput/AutocompleteList.tsx
```diff
@@ -246,16 +249,22 @@ export const AutocompleteList = React.forwardRef<HTMLDivElement, AutocompleteLis
...
-						height={`${listHeight}px`}
-						overscan={0}
-						itemContent={renderAutocompleteOption}
-						className='autocomplete-list__options'
-					/>
+					{isVirtuosoOpen && (
+						<Virtuoso
+							ref={virtuosoRef}
+							totalCount={list.length}
+							height={`100%`}
```
rzvdv, line 262: if you pass style={{ height: listHeight }} doesnt it solve the problem?

PR #511, opened 2023-01-13
Cursor is active after inputting session_alias in messages search filter and the list is not automatically shown

okti-web:


osinkinroman (commented):
I think we need to add a limit, so we can't add one session more than once

src/components/filter/row/MultipleStringFIlterRow.tsx
```diff
@@ -153,7 +156,9 @@ export default function MultipleStringFilterRow({ config }: MultipleStringFilter
 						value={config.currentValue}
 						setValue={config.setCurrentValue}
 						autoresize
-						autoCompleteList={config.autocompleteList}
+						autocompleteList={config.autocompleteList?.filter(
```
okti-web, line 159: should we change how this variable is called in the config? 
Because this line has the real "autocompleteList". Before it we just pass down the sessionsStore.booksStream, and use it for slightly different purposes (like validation in Bubble components), other than autocompletion

okti-web:
this contains tickets 4630, 4657, 4450

src/components/books/BookSelect.tsx
```diff
@@ -51,7 +51,7 @@ const BookSelect = () => {
 				onSubmit={onSubmit}
 				value={currentValue}
 				setValue={setCurrentValue}
-				autoCompleteList={booksIds}
+				autocompleteList={booksIds}
```
osinkinroman, line 54: I don't think its necessary to rename to autoComplete, because its one word "autocomplete"

src/components/util/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -108,10 +107,6 @@ const AutocompleteInput = React.forwardRef((props: Props, ref: any) => {
 			e.preventDefault();
 		}
 
-		if (value.trim().length > 0 && submitKeyCodes.includes(e.keyCode)) {
```
osinkinroman, line 111: you should return this, because submitKeyCodes contained KeyCodes.Tab , and right now we cant write name of needed session and press on TAB to submit it, as it was before

osinkinroman (changes requested):

src/components/filter/row/MultipleStringFIlterRow.tsx
osinkinroman: Just rename property of AutocompleteInput component "autoCompleteList" into "autocompleteList"

osinkinroman (approved):

PR #519, opened 2023-02-01
[TH2-4450] Duplicate session aliases and duplicated messages should not be shown

okti-web:
duplicate session aliases and duplicated messages should not be shown

### DIFF
--- a/src/stores/SearchStore.ts
+++ b/src/stores/SearchStore.ts
@@ -48,7 +48,7 @@ import notificationsStore from './NotificationsStore';
 import WorkspacesStore from './workspace/WorkspacesStore';
 import FiltersHistoryStore from './FiltersHistoryStore';
 import { SessionsStore } from './messages/SessionsStore';
-import { getItemAt } from '../helpers/array';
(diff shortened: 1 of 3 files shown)


src/stores/SearchStore.ts
```diff
@@ -486,6 +486,8 @@ export class SearchStore {
 		const bookId = this.booksStore.selectedBook.name;
 		const scope = this.searchForm.scope;
 
+		this.searchForm.stream = getArrayOfUniques(this.searchForm.stream);
```
osinkinroman, line 489: Not the best approach, I think its better to check for unic sessions when we just inserted session name into input, not when we initiate the search

osinkinroman (changes requested):

src/stores/SearchStore.ts
osinkinroman: Also I'm not sure if we can with the same approach insert, for example, a bunch of filters with the same name, maybe here we also need to check for unique values

osinkinroman (approved):

PR #532, opened 2023-02-14
th2-4710 autocomplete clicks don't work under certain conditions

okti-web:
autocomplete clicks don't work under certain conditions

1. Start typing a correct name, but stop in the middle and just click out of the input. 
2. get back to the created red bubble, try to choose an autocomplete option by clicking. 
3. fail to choose an option through clicking


rzvdv (changes requested):
commits and pr's should be created with a descriptive name

okti-web:
> commits and pr's should be created with a descriptive name

noted. So the only commit here moves setIsEditing from onBlur to the times when we switch bubbles. Somehow it starts working this way.

rzvdv (approved):

PR #539, opened 2023-02-16
TH2-4750   after we delete fix-client1->fix-client. we don't receive the current option as one of the autocomplete variants.

okti-web:


src/components/util/AutocompleteInput/AutocompleteList.tsx
```diff
@@ -70,7 +70,9 @@ export const AutocompleteList = React.forwardRef<HTMLDivElement, AutocompleteLis
 
 		const list = React.useMemo(() => {
 			if (!value || alwaysShow) return items;
-			return items.filter(item => item.toLowerCase().includes(value.toLowerCase()));
+			return items.filter(
+				item => item.toLowerCase().includes(value.toLowerCase()) && value !== item,
```
rzvdv, line 74: doesnt it solve the same issue you tried to solve by adding a 'closedOnClick' prop onAutocompleteInput, if so it should be deleted

src/components/filter/row/MultipleStringFIlterRow.tsx
```diff
@@ -44,6 +45,7 @@ export default function MultipleStringFilterRow({ config }: MultipleStringFilter
 
 	const valueBubbleOnChangeFor = (index: number) => (nextValue: string) => {
 		config.setValues(replaceByIndex(config.values, index, nextValue));
+		focusBubbleOrInput(index + 1);
```
rzvdv, line 48: What does it do?

okti-web: removed it. 
the idea was that it would allow us to target autocompleteInput after we press 'Tab', but it doesn't accomplish that.

src/components/util/AutocompleteInput/AutocompleteList.tsx
okti-web: I checked on the 4657 branch - this fix was for a different problem. We got rid of AutocompleteAnchor, so that it wouldn't re-select all the last sessions. 

This fix is made, to show all the autocomplete options. Because before it when we redact a Bubble, not all options showed up.

osinkinroman (commented):
In message panel, when we have for example 3 bubbles, then we edit second bubble and press 'tab' to submit, there is no focus on empty input. I think we have two options here: or we focus on the same bubble we edited before or focus empty input. Make what is simpler

src/components/search/SearchInput.tsx
```diff
@@ -55,10 +55,16 @@ export interface Props extends StateProps, DispatchProps {
...
 
+	private bubbleRefs: React.MutableRefObject<{
+		[index: number]: BubbleRef | null;
+	} | null> = React.createRef();
+
 	private root = React.createRef<HTMLDivElement>();
 
 	constructor(props: Props) {
 		super(props);
+		this.state = { isBubbleEditing: false, isReactFunctioning: true };
```
osinkinroman, line 66: delete unused isReactFunctioning

src/components/util/Bubble.tsx
```diff
@@ -27,25 +27,29 @@ interface Props {
 	style?: React.CSSProperties;
 	removeIconType?: 'default' | 'white';
 	value: string;
-	selectNext?: () => void;
-	selectPrev?: () => void;
+	setIsBubbleEditing: React.Dispatch<React.SetStateAction<boolean>>;
+	isBubbleEditing: boolean;
```
osinkinroman, line 31: delete unused isBubbleEditing

src/components/util/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -39,6 +39,8 @@ type Props = Override<
 	anchor?: HTMLElement;
 	autocompleteListMinWidth?: number;
 	closedOnClick?: boolean;
+	isBubbleEditing?: boolean;
+	outerInputRef?: React.MutableRefObject<HTMLInputElement | undefined>;
```
osinkinroman, line 43: delete unused outerInputRef

src/components/search/SearchInput.tsx
```diff
@@ -188,6 +258,21 @@ export class SearchInputBase extends React.PureComponent<Props> {
...
 
+		if (e.keyCode === KeyCodes.LEFT) {
+			if (this.bubbleRefs.current) {
+				e.preventDefault();
+				this.bubbleRefs.current[this.props.searchTokens.length - 1]?.focus();
+			}
+			return;
+		}
+
+		if (e.keyCode === KeyCodes.RIGHT) {
```
osinkinroman, line 269: this code does nothing, i guess delete it too

osinkinroman (changes requested):

osinkinroman (approved):
Also I think you need to add more description in ticket about changes in event's search input

okti-web:
> Also I think you need to add more description in ticket about changes in event's search input

done, thanks a lot!

PR #573, opened 2024-07-26
JSONL reader release

molotgor:


README.md
```diff
@@ -41,3 +41,46 @@ spec:
...
+- It has search which will highlight searched values in the viewing files. Current search could be exported in the JSON file, and imported from such files. Search files has this structure
+
+```
+[{
+  "pattern": "searched value",
+  "color": "hex color"
+}, ...]
+```
+
+- JSONL reader can also get data of Jupyter Notebooks using  `th2-json-stream-provider`, when notebook is requested it will be displayed as list of parameters which could be modified. Then this notebook could be launched and resulting JSONL file will be requested on finish and added to current list.
```
Nikita-Smirnov-Exactpro:
```suggestion
- JSONL reader can also generate data by Jupyter Notebooks execution using `th2-json-stream-provider`, when notebook is requested it will be displayed as list of parameters which could be modified. Then this notebook could be launched and resulting JSONL file will be requested on finish and added to current list.
```

README.md
```diff
@@ -41,3 +41,46 @@ spec:
...
+- Notebooks has default result of 1 in which case new results will overwrite previous result of this notebook, this could be changed in the bottom of notebook editing form, if amount is more than 1 then result will overwrite oldest one.
```
Nikita-Smirnov-Exactpro:
```suggestion
- Notebooks tab holds one one result by default in which case new results will overwrite previous result of this notebook, this could be changed in the bottom of notebook editing form, if amount is more than 1 then result will overwrite oldest one.
```

README.md
```diff
@@ -41,3 +41,46 @@ spec:
...
+- If notebook result had search file with structure stated before, it will replace current search.
```
Nikita-Smirnov-Exactpro, line 72: Please write information about notebook parameters responsible for result and search file.

README.md
```diff
@@ -41,3 +41,46 @@ spec:
...
 
 ![picture](screenshot.png)
+
+# JSONL reader
+
+## API
+
+This functionality needs `th2-json-stream-provider 0.0.7 (or newer)` backend component to function. It should be accessible at `{rpt-viewer-path}/json-stream-provider*`
+
+## Features
```
Nikita-Smirnov-Exactpro, line 51: Please add a separate topic with correct examples of each file:
* simple notebook with all kinds of parameters
* result with all kinds for key words
* search file
* parameter values

molotgor: Isn't this examples should be on provider read me?

README.md
molotgor: I think this information should be in provider readme, since this parameters don't interact with ui

Nikita-Smirnov-Exactpro: The detail information can be, but we should mentioned here about them.

README.md
Nikita-Smirnov-Exactpro: So we have tree entities:
* viewer which can work with specific file formats
* notebook which one nowhere described and can have any code
* js-p which handle request, pass some path to parameters, executes notebook and return result in any JSONL format.

I think detail for files are fully denuded on viewer

README.md
molotgor: This parameters aren't used in anything on ui, why it should be mentioned here?

